PR #43, opened 2025-04-20
feat: rss-feed

therobrob:
## Checklist

- [ ] Changed the date in updatet content pages 

- [ ] English
- [ ] German


therobrob:
Ich habe jetzt mal nur die News-Artikel mit aufgenommen. Soweit ich es verstanden habe, sollen die Seiten aufgenommen werden, die für deine Leser am interessantesten sind und regelmäßige Updates oder Inhalte bieten.

Vielleicht machen wir zukünftig noch eine Seite, wo wir Content-Updates posten? Oder sollten wir die unterste Ebene mit den Artikeln auch mit aufnehmen?

lenderom:
I'm not so sure to be honest. Because at the moment I'm not use about the use case or the problem we would like to solve. It could be fine to only hav news articles as part of the RSS Feed (maybe it's also worth to publish a whole news article for relevant changes in countries/operators). However it could also be okay, to have every change as part of the RSS Feed. But maybe this get's a bit messy to display every single litte change. And in the RSS Feed it wouldn't be possible to view this change. You would only get the whole page in the feed again and agian.

So yeah... probebly it's best to only have the news as part of the feed and publish a news article for relevant changes in the countries.

lenderom (approved):

therobrob:
yes.... it´s better than before i guess. We can improve this later on. I´ve added this in [Perspektive und Ideen](https://github.com/orgs/fipguide/projects/1/views/1?pane=issue&itemId=67196599) :)

PR #52, opened 2025-04-29
fix: Search overflows on large screens

MoritzWeber0:
## Description

The search overflowed the picture on larger screens. This should fix it.

Before:
image

After:
image

It's probably better to set the max-width on a parent element and let the search + picture inherit from it, but for some reason it didn't properly work when I tried it.


lenderom (approved):
Thank you for your bugfix  🫶

PR #96, opened 2025-05-03
feat: Use fontsource instead of restribution of fonts and icons

MoritzWeber0:
## Description

Use fontsource to install the fonts via npm instead of manually redistributing the fonts and icons. The benefit is that fonts can receive updates and icons don't have to be uploaded manually.

To use icons, you have to find out the name of the icon (https://fonts.google.com/icons). Then, there are two different ways: 
- As partial in html templates: 
  ```go
  {{ partial "icon" <icon-name> }}
  ```
- As shortcode in markdown templates: 
  ```go
  {{< icon <icon-name> >}}
  ```

TODO
- [ ] Replace all icons with the new approach
- [ ] Update the wiki


therobrob:
We've already used this way of icon and font integration in an earlier development stage. So we should discuss it with @lenderom.

MoritzWeber0:
> We've already used this way of icon and font integration in an earlier development stage. So we should discuss it with @lenderom.

@lenderom already wrote to me and I changed the PR to a draft for now. Let's discuss it in person 👍

MoritzWeber0:
@therobrob I changed it to rounded symbols 👍

assets/sass/expander.scss
```diff
@@ -34,18 +34,17 @@ details > summary::-webkit-details-marker {
...
   position: absolute;
   left: 1.2rem;
+  font-size: 24px;
+  @extend .material-symbols-rounded
 }
 
 details[open] > summary::before {
-  content: url($icon-remove);
+  content: "remove";
+  font-size: 24px;
```
therobrob, line 46:
```suggestion
  font-size: 2.4rem;
```

layouts/partials/icon.html
```diff
@@ -0,0 +1 @@
+<span class="material-symbols-rounded">{{ . }}</span>
```
therobrob, line 1:
```suggestion
<span class="material-symbols-rounded" aria-hidden="true">{{ . }}</span>
```
The icons are just decorative, so we should hide them for SR. There is always a label or aria-label to explain the link/button/and so on.

assets/sass/expander.scss
```diff
@@ -34,18 +34,17 @@ details > summary::-webkit-details-marker {
...
-  content: url($icon-add);
-  font-weight: normal;
-  width: 2.4rem;
-  height: 2.4rem;
-  margin-right: .8rem;
-  text-decoration: none;
+  content: "add";
   position: absolute;
   left: 1.2rem;
+  font-size: 24px;
```
therobrob, line 40:
```suggestion
  font-size: 2.4rem;
```

assets/sass/expander.scss
therobrob, line 46:
```suggestion
  font-size: 2.4rem;
```

therobrob (changes requested):

therobrob (approved):

Issue #95, opened 2025-05-02
Auto-generate change history of pages using Git log


MoritzWeber0:
Instead of the date that has to be changed manually for each page after editing, wouldn't it be cool to have an auto-generated change history for pages? This could potentially also lead to global update feed (related #43).

Issue #70, opened 2025-05-01
SEO / Google Optimization


MoritzWeber0:
There are some tasks for optimizing google ranking & seo in general.
Split up in sub-issues

Originally reported by @therobrob

therobrob:
Questions:
1. are our "country" and "operator" articles of type "blog" or something like that? should we use structures data for it?
2. structured data for NewsArticle provides "datePublished" AND "dateModified" – do we want to use that in md-Files?
3. structured data for NewsArticle uses images. For now, we have an image for each md-File in news-folder.
4. in news article we have the "author" attribute. Do we want to use it for structured data? For now, i use "FIP Guide" as an organization in google author-attribute.

therobrob:
1. type article
2. we wont show datePublished, only dateModified
3. one image for now :)
4. "FIP Guide" is ok

PR #98, opened 2025-05-03
feat: Auto-fetch "last updated" from Git history

MoritzWeber0:
## Description

It's not longer needed to provide the "date" attribute in the page metadata. If it's not provided, the information will be auto-fetched from Git.
Related to #95

The disadvantage with the auto-updated date is that also minor changes like updated grammar, spelling etc. will also update the "last updated" field.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,6 @@
 ## Checklist
 <!-- Check fields with: [x] / Abhaken von Punkten: [x] -->
 
-- [ ] Changed the date in updatet content pages <!-- Auf Inhaltsseiten wurde das Bearbeitungsdatum angepasst -->
 - [ ] Check the License of new pictures (non-commercial use without attribution) <!-- Die Lizenz neuer Bilder geprüft (nicht-kommerzielle Nutzung ohne Namensnennung) -->
 
 The content was modified in the following languages: <!-- Der Inhalt wurde für die folgenden Sprachen angepasst -->
--- a/archetypes/country.de.md
+++ b/archetypes/country.de.md
@@ -1,5 +1,4 @@
 ---
-date: '{{ .Date }}'
 draft: <true/false>
 title: '<Name des Landes in Deutsch>'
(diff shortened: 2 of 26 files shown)


lenderom:
In my personal opinion this is a pretty neat feature. However I think we should discuss this topic.
I think it's pretty good for not forgetting to update the date.

However I don't know if we would always like to display the authors name? Mayber we should show the name, when hovering over the infos, as well as a link to the last commit/change of this page using the .Hash variable to build a URL.

MoritzWeber0:
> In my personal opinion this is a pretty neat feature. However I think we should discuss this topic. I think it's pretty good for not forgetting to update the date.
> 
> However I don't know if we would always like to display the authors name? Mayber we should show the name, when hovering over the infos, as well as a link to the last commit/change of this page using the .Hash variable to build a URL.

I can move the name to the tooltip, that's no issue and probably makes sense. And a link should be fine too, but not sure where to place it.

therobrob:
The author's information is also relevant for Google, so we should decide it. See https://github.com/fipguide/fipguide.github.io/issues/70#issuecomment-2847786685.

Unfortunately, the tooltip feature is currently not available via screen reader – could you please adjust this?

MoritzWeber0:
> Unfortunately, the tooltip feature is currently not available via screen reader – could you please adjust this?

Made it keyboard usable and should be recognized now. Let me know if there is anything else to improve from accessibility perspective :) 
image

lenderom:
- accessible
- link to file history in github
- no hover
- no name only date

lenderom:
I just removed the "date" metadata from the wiki

MoritzWeber0:
> * accessible
> 
>     * link to file history in github
> 
>     * no hover
> 
>     * no name only date

The proposed changes are implemented now.

MoritzWeber0:
The "Updated at" is now also pagefind-ignored and no longer shows up in the search :)

lenderom (approved):

Issue #80, opened 2025-05-02
Add QA steps to the pipeline


therobrob:
To improve the delivery quality, we should consider adding QA steps to the pipeline; e.g.

- check for code-style conventions 
- AXE for a11y
- check whether all languages ​​have been taken into account when making content changes

MoritzWeber0:
Some parts could also be checked in a pre-commit hook. Maybe it also makes sense to check some Markdown formatters.

PR #99, opened 2025-05-03
ci: Add pre-commit hooks for basic formatting + gitleaks

MoritzWeber0:
## Description

Add pre-commit hooks: 

- Gitleaks to identify secrets before pushing them
- Check for large files (to avoid a large repository)
- Check for invalid file structures
- Add new line to the end of each file (to avoid changes to the last line when content is added to a file)
- Remove trailing whitespaces. Since sometimes a double whitespace was used for a new line, I changed it to `\` for new lines. This is a lot more readable (whitespaces are usually not displayed in default editor setups)
- actionlint to check for valid Github Actions files

See https://github.com/fipguide/fipguide.github.io/issues/80


therobrob:
Sounds good.
Two Questions:
- what´s the limitation for the file size? Do we need to specify a value here?
- Is it mandatory to use a \ for a newline? I think it's a bit counterintuitive for contributors, isn´t it?

MoritzWeber0:
>  what´s the limitation for the file size? Do we need to specify a value here?

The default is 500kB, but we can also change it to a different value.

>  Is it mandatory to use a \ for a newline? I think it's a bit counterintuitive for contributors, isn´t it?

That's the Markdown syntax, either `<br />`, two new lines, `\` or two whitespaces. There are reasons to avoid the two whitespaces because it's hard to read (as mentioned above, default editors don't show trailing whitespaces, and in reviews it's also hard to overlook them). 

To be fair, I don't expect non-tech contributors to run the `pre-commit` hooks, it's also not part of the content contribution guide. That's something we as maintainers have to do in the end.

PR #100, opened 2025-05-03
fix: Reduce margins for anchors

MoritzWeber0:
## Description

This is just a proposal because in my opinion the margins of the anchors / dividers are too large. But it's an personal opinion, so happy to hear other voices :) 

Before: 
image

After: 
image


therobrob:
I think it's a personal preference :) I'm a fan of airier layouts – in addition to the separator element, larger spacing visually indicates that this is a new chapter.

PR #130, opened 2025-05-08
fix: improve performance by using caching

lenderom:
## Description

Improve the build performance by using partial caches

## Checklist

- [x] English
- [x] German


lenderom:
The most significant improvement would be, if we achieve to do the image processing only once per image. However we can't use partialCached for this, because this is still done once per language. So each image is processing twice at the moment, even if it's the same image (we don't have different images for different languages). Maybe @MoritzWeber0 has an idea?

MoritzWeber0:
> The most significant improvement would be, if we achieve to do the image processing only once per image. However we can't use partialCached for this, because this is still done once per language. So each image is processing twice at the moment, even if it's the same image (we don't have different images for different languages). Maybe @MoritzWeber0 has an idea?

The image processing is automatically cached and only done once per image - prove me wrong. The detail lies in the sentence of the last paragraph: "Hugo caches processed images in the resources directory.". When I have a look at the resources directory, I only see general images and images in the "en" directory. There is no additional "de" directory, so I'd assume that these images are no rerendered.
image

lenderom:
No I mean during the build process.

When running `hugo server --ignoreCache --templateMetrics --templateMetricsHints` you can see how often the partial image.html is called and how often it is cached.

Every Country Flag (4)
And every News Article (3)
And the Startpage image (1)

And all of this for each language.

MoritzWeber0:
> No I mean during the build process.
> 
> When running `hugo server --ignoreCache --templateMetrics --templateMetricsHints` you can see how often the partial image.html is called and how often it is cached.
> 
> Every Country Flag (4)
> And every News Article (3)
> And the Startpage image (1)
> 
> And all of this for each language.

That makes sense. The partial is called for each language. The question is: Is the image processing inside the partial cached? 

So if I call the same image partial twice, it might resize the images the first time and the second time the expensive resizing step *might be* skipped and it's faster. 

I'll have to look into at another time. 

(In the worst case we can do the processing on our own and then ship the images via the static dir)

lenderom:
Resizing is nativly only done once by hugo.
But we are using runner caches for the resource section for hugo now, which make a huge impacht in build speed:

Before:
image

After:
image

PR #140, opened 2025-05-18
refactor: Replace "ico" partial with "logo" partial

MoritzWeber0:
The icon usage was replaced by fontsource in #96. The remaining icon infrastructure was only used for the logos of operators.

This commit simplifies the code by replacing the "ico" partial with a "logo" partial.

## Checklist

- [ ] English
- [ ] German


lenderom:
I thought we need the icon partial for custom icons in the future. Like "Reservation Mandatory" or other train specific situations, where no default icons exist. :O

MoritzWeber0:
> I moved the SVG Logos from the svg folder to the images folder. I think it's confusing to separate between these two folders (I mean we had SVGs in the images folder before :D )

At some point, when we have more coverage of operators, it might make sense to move the logos to the content folders, similar to the content images.

PR #145, opened 2025-05-24
fix: Remove padding of search highlight and unify bg-color

### LINKED ISSUE #132: Update css highlight class
https://www.fipguide.org/de/operator/eurostar/?highlight=euro
Image

The highlighting has an unpleasant look

MoritzWeber0:
Resolves #132


therobrob (commented):
Do we really need the important flag?

MoritzWeber0:
> Do we really need the important flag?

The padding is part of the pagefind styles and I didn't find a good way to overwrite it in another way.

MoritzWeber0:
> > Do we really need the important flag?
> 
> The padding is part of the pagefind styles and I didn't find a good way to overwrite it in another way.

Nevermind, I found a better way using a class with higher CSS specificity.

therobrob (approved):

Issue #104, opened 2025-05-04
Explanation for highlight-snippet


therobrob:
For now, there is no explanation what the color-codes in the highlight-snippet mean.
We should add an headline with the type.

**Example: "Persönlicher Tipp":**
<img width="857" alt="Image" src="https://github.com/user-attachments/assets/f2cb6174-64ed-48c3-bd18-6e7f1e26ccfd" />

therobrob:
which version is your favorite? :)
i don´t want to use a heading for it, because in most of the cases there already is an heading outside the highlight box.

<img width="861" alt="Image" src="https://github.com/user-attachments/assets/e1c87c98-4b36-4db7-a897-7e13a73dc041" />
<img width="862" alt="Image" src="https://github.com/user-attachments/assets/47e5289d-435e-45fd-9deb-45571da5461b" />

therobrob:
Good idea, i‘ll add an Icon :)

PR #149, opened 2025-05-24
feat: explanation & styling for highlight

therobrob:
## Description

[#104](https://github.com/fipguide/fipguide.github.io/issues/104)

## Checklist

- [ ] English
- [ ] German

### DIFF
--- a/layouts/shortcodes/highlight.html
+++ b/layouts/shortcodes/highlight.html
@@ -1,3 +1,4 @@
-<div class="m-text-highlight m-text-highlight--{{ (.Get 0) | markdownify }}" aria-label='{{ T (printf "highlight-%s" (.Get 0)) }}'>
-    {{ .Inner }}
+<div class="m-text-highlight m-text-highlight--{{ (.Get 0) | markdownify }}">
+    <div class="m-text-highlight__roofline"> {{ T (printf "highlight-%s" (.Get 0)) }} </div>
+    <div>{{ .Inner }}</div>
 </div>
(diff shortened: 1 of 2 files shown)


layouts/shortcodes/highlight.html
```diff
@@ -1,3 +1,4 @@
-<div class="m-text-highlight m-text-highlight--{{ (.Get 0) | markdownify }}" aria-label='{{ T (printf "highlight-%s" (.Get 0)) }}'>
-    {{ .Inner }}
+<div class="m-text-highlight m-text-highlight--{{ (.Get 0) | markdownify }}">
```
MoritzWeber0, line 1:
```suggestion
<div class="m-text-highlight m-text-highlight--{{ .Get 0 }}">
```

I know that you didn't touch it in this PR, but the argument is just a simple string and not in markdown format. So there is no need to markdownify it.

layouts/shortcodes/highlight.html
```diff
@@ -1,3 +1,4 @@
...
+    <div class="m-text-highlight__roofline"> {{ T (printf "highlight-%s" (.Get 0)) }} </div>
```
MoritzWeber0, line 2:
```suggestion
    <div class="m-text-highlight__roofline"> {{ T (printf "highlight.%s" (.Get 0)) }} </div>
```

Since the i18n files are yaml and they support nesting, I'd make use of it for a simpler file structure. Example with the `en.yaml`:

```
highlight:
  important: Important Information
  inofficial: Unofficial Information
  tip: Personal Tip
```

PR #151, opened 2025-05-25
feat: Add sitemap to robots.txt

MoritzWeber0:
Make it easier for crawlers to find the sitemap.

I was wondering why in the Google Search Console only 16 of the 52 pages were crawled.
One potential reason is that it can't find the sitemap.xml as it's not part of the `robots.txt`.

I added the sitemap manually in the search console, but we should make all crawlers aware of the sitemap.


therobrob (approved):
Good job, thx!

PR #159, opened 2025-05-27
fix: Detection of image in pagefind was broken

MoritzWeber0:
`data-pagefind-meta="img"` is not a valid option, therefore it was ignored and the first picture of the page was used.


lenderom (approved):
Good find! 😊

PR #164, opened 2025-05-28
fix: create a working prettier config

lenderom:
## Description

Make some files better readable by adding a working prettier config. Only merge after #163 has been merged.

## Checklist

- [x] English
- [x] German


lenderom:
It seems like something broke with the icon partial which is used in the sidemenu.

MoritzWeber0:
Before merging this, we should make sure that all open PRs are merged. It will to lead to many conflicts otherwise.

MoritzWeber0:
Superseded by #331

PR #177, opened 2025-06-03
fix: Always place the copy icon on the right side on mobile

### LINKED ISSUE #49: Copy-Icon Mobile-View
Image

Bei mehrzeiligen Überschriften rutscht das Iocn ganz rechts an den Rand. Das sieht eher suboptimal aus.

MoritzWeber0:
Instead of having the copy icon directly after the text (which produced some issues with multi-line texts), always place the the copy icon on the right side. It's easier to reach for right-handed people.

Not 100% sure if the link icon is still recognized with a connection to the title, so happy to hear any feedback.

Resolves #49

### DIFF
--- a/assets/sass/anchorlink.scss
+++ b/assets/sass/anchorlink.scss
@@ -1,9 +1,13 @@
 .a-anchorlink {
   display: flex;
-  flex-wrap: wrap;
+  flex-wrap: nowrap;
   margin-bottom: 1.2rem;
   align-items: center;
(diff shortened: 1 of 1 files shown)


therobrob:
I can understand very well why you changed it. I also find the current placement of the button annoying when it wraps in new line. However, I think the spacing is too large for a short headline or a larger screen width ._.

MoritzWeber0:
It would be easier to place it left to the text: 

image

PR #180, opened 2025-06-06
feat: Present booking methods as custom expanders

MoritzWeber0:
Move booking information to own pages. Embed the pages via custom expanders to the operator pages.

PR #189, opened 2025-06-09
fix: Mark `tel:` links as safe

### LINKED ISSUE #185: Phone links don't work
The linked phone numbers don't work. Instead of the `tel#` href, the href is replaced with something else.

Reproduction steps: 

1. Open https://www.fipguide.org/en/operator/eurostar
2. Click on a phone number

zusorio:
## Description
By default, Hugo only allows using `http:`, `https:`, and `mailto:`. URLs using other schemes are replaced with `#ZgotmplZ`. To allow using `tel:`, we have to declare it as safe using `| safeURL`. This marks every URL as safe which should be fine, since all content inside the repo is trusted.
See https://gohugo.io/functions/safe/url/ for more info.

Closes #185

## Checklist

- [ ] English
- [ ] German

### DIFF
--- a/layouts/_default/_markup/render-link.html
+++ b/layouts/_default/_markup/render-link.html
@@ -2,5 +2,5 @@
 {{- if strings.HasPrefix $url "http" -}}
 {{ partial "link" . }}
 {{- else -}}
-<a href="{{ $url }}">{{ .Text }}</a>
+<a href="{{ $url | safeURL }}">{{ .Text }}</a>
 {{- end -}}


lenderom (approved):
Hi Tobi,
Good catch, thanks for your contribution.
I think the first use case will be in #180, so let's merge it there :)

MoritzWeber0:
> Hi Tobi, Good catch, thanks for your contribution. I think the first use case will be in #180, so let's merge it there :)

The `tel:` links are already used on the Eurostar website, so there is an immediate improvement :)
Thanks @zusorio!

Issue #77, opened 2025-05-02
A11Y-Improvements


therobrob:
We've taken accessibility into account from the very beginning. However, some things need to be reviewed and improved:
- Check and standardize focus indicators
- Check heading hierarchy
- Check usability with screen reader (at best NVDA & VoiceOver)
- Is the `<hr>` element captured by the screen reader? If so, we should hide it
- Can we improve the SR-support in the markdown-pages? e. g. [does role="doc-endnotes" need title / aria-label / aria-labelledby](https://kb.daisy.org/publishing/docs/html/dpub-aria/doc-endnotes.html#usage)? The marks of ol-elements are currently read by SR
- Within the expander, there are `<br>` that are read as "Empty Object." Can we remove them?

PR #192, opened 2025-06-12
feat: Implement enhancements for new design

MoritzWeber0:
I tried to split the different parts into different commits, so that we can easily take out some individual parts.

### DIFF
--- a/assets/sass/navigation.scss
+++ b/assets/sass/navigation.scss
@@ -57,7 +57,7 @@
 
 .o-header__logo {
   display: flex;
-  margin: .6rem;
+  margin: .6rem 0;
   position: relative;
   text-decoration: none;
   border-radius: var(--border-radius-m);
--- a/assets/sass/navigation.scss
+++ b/assets/sass/navigation.scss
@@ -14,6 +14,7 @@
 .o-header__wrapper {
   display: flex;
   justify-content: space-between;
+  padding: 0;
 
   @media (min-width: #{$breakpoint-md}) {
     justify-content: unset;
@@ -58,7 +59,6 @@
   display: flex;
   margin: .6rem;
   position: relative;
-  left: -1.6rem;
   text-decoration: none;
   border-radius: var(--border-radius-m);
 
--- a/assets/sass/styles.scss
+++ b/assets/sass/styles.scss
@@ -212,4 +212,16 @@ img {
   > div {
     align-content: center;
   }
+
+  @media (max-width: #{$breakpoint-md}) {
+    h2 {
+      font-size: 4.8rem;
+    }
+  }
+}
+
+.o-startpage__image {
+  @media (max-width: #{$breakpoint-md}) {
+    display: none;
+  }
 }
--- a/assets/sass/styles.scss
+++ b/assets/sass/styles.scss
@@ -196,7 +196,7 @@ img {
 }
 
 .o-startpage .row {
-  margin-bottom: 3.2rem;
+  margin-bottom: 2.4rem;
 }
 
 .o-startpage__intro {
--- a/assets/sass/styles.scss
+++ b/assets/sass/styles.scss
@@ -9,7 +9,7 @@ body {
   overflow-x: hidden;
 }
 
-a:not(.m-teaser) {
+a {
   color: var(--link-default);
   transition: all 0.3s ease;
   text-underline-offset: .2rem;
--- a/assets/sass/teaser.scss
+++ b/assets/sass/teaser.scss
@@ -82,3 +82,11 @@
 .m-teaser__content {
     padding: .8rem;
 }
+
+.m-teaser__text {
+    line-clamp: 3;
+    display: -webkit-box;
+    -webkit-line-clamp: 3;
+    -webkit-box-orient: vertical;
+    overflow: hidden;
+}
--- a/assets/sass/teaser.scss
+++ b/assets/sass/teaser.scss
@@ -20,11 +20,6 @@
 
 .m-teaser--listview .m-teaser__wrapper {
     margin-bottom: 2.4rem;
-    flex-direction: row;
-
-    @media (max-width: #{$breakpoint-md}) {
-        flex-direction: column;
-    }
 
     .m-teaser__image {
         @media (min-width: #{$breakpoint-md}) {
@@ -46,8 +41,7 @@
     .m-teaser__wrapper {
         width: 100%;
         height: 12rem;
-        padding: .8rem;
-        flex-direction: row;
+        padding: 1.4rem;
         justify-content: space-between;
 
         span {
@@ -59,9 +53,20 @@
 
     a {
         h3 {
-            text-decoration: underline;
-            align-self: center;
+            overflow-wrap: normal;
+            word-break: normal;
+            white-space: normal;
+            hyphens: manual;
             margin-bottom: 0;
+        }
+
+        transition: background 0.2s, color 0.2s;
+
+        &:hover,
+        &:focus {
+            background: var(--link-hovered);
+            border-color: var(--link-hovered);
+            color: #fff;
         }
     }
 }
--- a/assets/sass/teaser.scss
+++ b/assets/sass/teaser.scss
@@ -1,10 +1,10 @@
 .m-teaser {
     display: flex;
+    text-decoration: none;
 }
 
 .m-teaser__wrapper {
     text-decoration: none;
-    color: var(--link-special);
     background-color: var(--color-onDark);
     box-shadow: var(--box-shadow);
     border-radius: var(--border-radius-l);
@@ -15,6 +15,20 @@
         @media (max-width: #{$breakpoint-lg}) {
             margin-bottom: 2rem;
         }
+    }
+
+    &:hover,
+    &:focus {
+        background: rgba($link-hovered, 0.15);
+
+        .m-teaser__headline {
+            text-decoration: underline;
+        }
+    }
+
+    .material-symbols-rounded {
+        font-size: 2.4rem;
+        margin-top: 1rem;
     }
 }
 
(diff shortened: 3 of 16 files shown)


assets/sass/navigation.scss
```diff
@@ -56,9 +57,8 @@
 
 .o-header__logo {
   display: flex;
-  margin: .6rem;
+  margin: .6rem 0;
   position: relative;
-  left: -1.6rem;
```
therobrob, line 61: i used it to align the logo to the left content border. do you think the logo should breach from it?

assets/sass/styles.scss
```diff
@@ -180,11 +180,20 @@ img {
 }
 
 .o-startpage .row {
-  margin-bottom: 3.2rem;
+  margin-bottom: 2.5rem;
```
therobrob, line 183: If you prefer a smaller distance, fine. But we should stick to steps of four.
therefore: 2.4rem

assets/sass/styles.scss
```diff
@@ -180,11 +180,20 @@ img {
...
   text-wrap: balance;
+  margin-bottom: 2rem;
+  display: grid;
+  grid-template-columns: 1fr 1fr;
+  gap: 2rem;
+
+  @media (max-width: #{$breakpoint-md}) {
+    display: flex;
+    flex-direction: column;
+  }
```
therobrob, line 196: What added value does it have compared to Bootstrap?

assets/sass/styles.scss
```diff
@@ -196,4 +205,47 @@ img {
...
+  @media (max-width: #{$breakpoint-md}) {
+    h2 {
+      font-size: 4.8rem;
+    }
+  }
+}
+
+.o-startpage__image {
+  @media (max-width: #{$breakpoint-md}) {
+    display: none;
```
therobrob, line 218: Why do you remove the image in small breakpoints? If you think it's unnecessary, why do we show it in large breakpoints?

assets/sass/styles.scss
```diff
@@ -196,4 +205,47 @@ img {
...
+
+.o-startpage__news-teaser-wrapper {
+  display: grid;
+  grid-template-columns: 1fr 1fr 1fr;
+  gap: 2rem;
+
+  @media (max-width: #{$breakpoint-md}) {
+    display: flex;
+    flex-direction: column;
+  }
```
therobrob, line 249: What added value does it have compared to Bootstrap?

assets/sass/teaser.scss
```diff
@@ -10,44 +10,35 @@
     border-radius: var(--border-radius-l);
     display: flex;
     flex-direction: column;
-
-    &:nth-child(-n+3) {
-        @media (max-width: #{$breakpoint-lg}) {
-            margin-bottom: 2rem;
-        }
-    }
```
therobrob, line 18: It seems very narrow to me now. Don't you like the change?

assets/sass/teaser.scss
```diff
@@ -59,16 +50,29 @@
...
         }
+
+        transition: background 0.2s, color 0.2s;
+
+        &:hover,
+        &:focus {
+            background: var(--link-hovered);
+            border-color: var(--link-hovered);
+            color: #fff;
+        }
```
therobrob, line 67: 1. the color change is too heavy in my opinion. 
2. It's important to me that we have consistency. In my version, the headline is underlined. If we want to change that, we should apply it to all teasers, not just the shortcut tiles.

assets/sass/teaser.scss
```diff
@@ -59,16 +50,29 @@
...
     }
 }
 
 .m-teaser__image img {
     aspect-ratio: 16/9;
     border-radius: var(--border-radius-l) var(--border-radius-l) 0 0;
+    object-fit: cover;
+    height: 100%;
```
therobrob, line 75: in my opinion, we should show the images always in aspect-ratio 16/9. Don't you think so?

assets/sass/styles.scss
```diff
@@ -196,4 +205,47 @@ img {
...
+    display: grid;
+    grid-template-columns: 1fr 1fr 1fr 3fr;
+    gap: 2rem;
+    margin-bottom: 2rem;
+
+    @media (max-width: #{$breakpoint-md}) {
+      display: flex;
+      flex-wrap: wrap;
+      gap: 1rem;
+    }
```
therobrob, line 232: In narrower views the boxes seem inharmonious to me. What do you think about going to `display: flex` in LG?
Bildschirmfoto 2025-06-14 um 12 22 34

assets/sass/teaser.scss
```diff
@@ -80,5 +84,13 @@
 }
 
 .m-teaser__content {
-    padding: .8rem;
+    padding: 1.4rem;
+}
+
+.m-teaser__text {
+    line-clamp: 3;
```
therobrob, line 91: in my option, three rows are too much

layouts/_default/home.html
```diff
@@ -1,39 +1,37 @@
...
 <section class="container">
-    <div class="row justify-content-center">
-        <h2>{{ T "news-headline"}}</h2>
-        {{ range first 3 (where site.RegularPages "Section" "news") }}
-          {{ partial "teaser.html" (dict "listview" false "type" "news" "page" .) }}
-        {{ end }}
+    <h2>{{ T "news-headline"}}</h2>
+    <div class="o-startpage__news-teaser-wrapper">
+      {{ range first 3 (where site.RegularPages "Section" "news") }}
+        {{ partial "teaser" (dict "listview" false "type" "news" "page" .) }}
```
therobrob, line 33: why do you removed "html" here, but not in line 10 for example? shouldn't we are consistent with it? 🤔

layouts/partials/teaser.html
```diff
@@ -22,7 +22,7 @@ <h3 class="m-teaser__headline">{{ $page.Title }}</h3>
       </a>
 
       <div class="m-teaser__text">
-        {{ $page.Summary | plainify | truncate 100 }}
+        {{ partial "strip-material-icons" $page.Summary | plainify | truncate 500 }}
```
therobrob, line 25: Do we need truncate if we use a line clamp?

therobrob (commented):
i've some questions and comments about your suggestions. to resolve them, it seems cool to talk about it, i think :)

assets/sass/styles.scss
MoritzWeber0: On small mobile devices (I took the iPhone SE as example), the image used 50% of the screen space, the title another 50%. The search wasn't visible anymore. I'd consider the image as decorative element that we can show on desktop devices where we have enough space available, but should focus on the important information on mobile devices. Check the screenshot, this is how it looks with the image: 

image

assets/sass/styles.scss
MoritzWeber0: Some as here: https://github.com/fipguide/fipguide.github.io/pull/192#discussion_r2150556246

assets/sass/navigation.scss
MoritzWeber0: I did some refactoring: The manual relative positioning was compensating margins that have been added to the `.o-header__logo` and `.container`. Instead, I just removed the margins and the manual positioning. 
When enabling the ruler in the Chrome DevTools, the alignment of the logo matches the alignment of the content:
image

In the previous version it was slightly misaligned (but only minimally, that would not have been noticed and wasn#t the intention of the change): 
image

assets/sass/teaser.scss
MoritzWeber0: image

Do you mean this margin? I found the margin of the previous version too large, but that's personal preference.

Previous version: 
image

layouts/_default/home.html
MoritzWeber0: There are a bunch of places where the html suffix is still there. I removed the html prefix everywhere in a new commit.

layouts/partials/teaser.html
MoritzWeber0: Yes. Line clamp is just a CSS property and doesn't influence the size of the DOM. With truncate, only a part of the content is added to the DOM of the list view. If we don't use it, it would always add the full content, which increases the size of the page slightly.

assets/sass/styles.scss
MoritzWeber0: Changing breakpoint-md to breakpoint-lg does it.

assets/sass/teaser.scss
therobrob: These are two separate aspects of it:

1. Using 16/9 aspect ratio
2. Prevention of whitespace

Do you think, there is no way / no need to keep 16/9? Otherwise we lose control over the shown part of the image :)

assets/sass/teaser.scss
therobrob: My concerns are mostly about accessibility and usability. The teasers must look clickable and have to be accessible. What do you think of this suggestion:
1. All teasers use the hover effect of the buttons in #180
2. To get a clickable look all teasers use the arrow icon, in the bottom left corner (like bahnhof.de for example)
3. in default no teaser headline is underlined, just on hover/focus (like the expanders)
4. on hover the whole card gets the hover effect (with ::after Element of the link)
5. on focus the headline is becoming the focus indicator

assets/sass/styles.scss
MoritzWeber0: I removed it from this PR and added it to https://github.com/fipguide/fipguide.github.io/pull/208.

assets/sass/teaser.scss
MoritzWeber0: 1-3 is implemented.
4: Not sure what you mean with the `::after` element of the link.
5: Is this possible without JS?

therobrob: thank you, looks good :)
4 + 5: the effect can be realized with the following code example:
`<div class="teaser"><div class="teaser-image"></div><div class="teaser-content"><time><h3 class="teaser_title"><a href="/link"><span>Überschrift</span>::after</a></h3><div class="teaser-text">Teaser-Text</div></div></div>`

The `::after`-Element has the size of the teaser and controls the hover & focus state.

I like the actual implementation, so i've added this to Issue https://github.com/fipguide/fipguide.github.io/issues/77, where some more a11y-feature could be added :)

PR #247, opened 2025-07-17
feat: Rotate over cities in footer-love

MoritzWeber0:
Rotate cities in the footer by blending in and out the cities. This allows us to add more cities like Paris in the future without taking much more space.

It's an experiment since it's the first animation of this kind in the FIPGuide. I'd like to hear some opinions about it, is it too distracting? 

Hugo doesn't support arrays / list in translation files, so I had come up with a custom solution for it by enumerating the entries.

### DIFF
--- a/assets/js/cityRotator.js
+++ b/assets/js/cityRotator.js
@@ -1,0 +1,29 @@
+(function() {
+  function initializeCityRotator() {
+    const rotatorElement = document.querySelector('.o-footer__city-rotator');
+    if (!rotatorElement) return;
+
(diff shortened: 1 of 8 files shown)


therobrob:
i like the animation! For me it is dynamic, but not distracting. For releasing the feature, it is sufficient to add the following aspects for a11y-reasons: 
- use of "reduced motion"
- option to play/pause this endless animation ([WCAG SC 2.2.2](https://www.w3.org/WAI/WCAG22/Understanding/pause-stop-hide.html))

PR #248, opened 2025-07-19
fix: Avoid layout shifts by pre-allocating space for pictures

MoritzWeber0:
So far, images don't set a fixed height for images. This leads to layout shifts since CSS can only recognize the size of the image when loaded.

By setting the aspect ratio on the `img` element, the browser can already pre-allocate the necessary space. When the image is loaded, the layout no longer shifts.

Layout shifts are one of the main complaints in the Google Lighthouse report: 
image


therobrob:
Hi @MoritzWeber0, i've noticed the error in footer.hmtl already. There were some email-notifications about it and i already fixed it. There is no need to fix this in this PR. :)

therobrob (approved):

PR #252, opened 2025-07-21
fix: Add aspect ratios for logo, operator logos and country flags

MoritzWeber0:


layouts/operator/list.html
```diff
@@ -7,7 +7,7 @@ <h1 data-pagefind-meta="title">{{ .Title }}</h1>
     <a href="{{ .RelPermalink }}" class="o-list__link">
       <div class="o-list__picture">
         {{ $object := replace .Path (printf "/%s/" .Page.Type) "" }}
```
therobrob, line 9: do we need `$object` anymore?

therobrob (changes requested):

layouts/operator/list.html
MoritzWeber0: Good catch, it's no longer needed.

therobrob (approved):

PR #267, opened 2025-07-28
feat: Improve dark mode

therobrob:
I've done some improvements for dark mode. Boxes and Cards now stand out more from the background. The logo and intersectors now react on color change.


layouts/partials/intersector.html
MoritzWeber0, line 1: What's the rationale behind embedding the intersector directly? When using directly, the SVG is part of the main content, increases the main content page, can't be lazy loaded and is not cached by the browser. In this case, the page size increases from 5.4kB to 8,5kB. That's still small enough, but if there is no explicit reason, I'd prefer the old solution.

layouts/partials/header-logo.html
MoritzWeber0, line 1: See the comment about the intersectors. The same argumentation applies here. 

https://github.com/fipguide/fipguide.github.io/pull/267/files#r2246201262

assets/sass/_variables.scss
```diff
@@ -61,4 +63,5 @@ html[data-theme="dark"] {
   --highlight-color-tip: #1A4A5C;
   --highlight-color-inofficial: #2A2D30;
   --highlight-color-important: #4A2A1A;
+  --border: .1rem solid #3d444d;
```
MoritzWeber0, line 66: What do you think of removing the border? The border with the color is quite noticeable. If the contrast is not good enough, maybe we can use a slightly darker bg-neutral and a slightly lighter bg-default?

If we decide to keep the border, it should also be applied to the operator and country overview pages.

MoritzWeber0 (commented):
Good improvement, I like the new color scheme. Thanks for adjusting the logo, that's definitely a part that was missing. We should rethink the way the logo and dividers are embedded, more about it in the comment section.

therobrob:
My work seems not to be good enough for you, so I’ve closed the PR.

MoritzWeber0 (approved):

PR #271, opened 2025-07-30
fix: Preserve language on logo-click and remove unused menu

MoritzWeber0:


MoritzWeber0:
> did you see PR 267? menu.html is removed there already..

Just saw it. I can remove the deletion from this PR.

PR #280, opened 2025-08-20
feat: add portugal & cp

therobrob:
## Description

Added portugal as country
Added CP as operator
Modified archtetype description

## Checklist

- [x] English
- [x] German
- [x] French


content/operator/cp/index.de.md
```diff
@@ -0,0 +1,164 @@
...
+
+{{% expander "Alfa Pendular (AP) ⚠️" traincategory "category" %}}
+**Beschreibung:** \
+Schnelle Neigetechnik-Züge (Pendolino), die auf der Strecke zwischen Braga, Porto, Lissabon und Faro mit wenigen Halten verkehren. Die Züge bieten einen hohen Komfort, Wi-Fi, Steckdosen an jedem Platz sowie einen Cafeteria- und Barservice.
+
+FIP Freifahrtsscheine (egal welcher Klasse) sind nur in der zweiten Klasse gültig.
+
+**Reservierung möglich:** Ja \
+**Reservierungspflicht:** ⚠️ Ja \
+**Kosten für Reservierung:** 5€
```
lenderom, line 47: Do you think we should add the reservation price to the booking channels? Or is this reservations price specific for this train category?

therobrob:
> > ### questions
> > 
> > 1. Celta Train runs between Portugal and Spain in cooperation of Renfe and CP. Should we add Renfe to Portugal and CP to Spain or is it an edge case?
> > 2. some border points aren't in use anymore – should we show these ones?
> > 3. Is there always FIP 50 / Globalpreis?
> 
> @therobrob
> 
> 1. I think it's enough if we only add the operator to it's own country. I think the Celta is not much different the other EC trains like thes in Germany in cooperation between DB and CD/PKP/DSB/ÖBB/SBB... I guess it could be confusing to have both operators in both countries.
> 2. I think we should add them anyway maybe they are again in use in the future.
> 3. You mean in the info section?

thanks for answering, your commits cover all three questions :)

content/operator/cp/index.de.md
therobrob: it is not specific for this train category, but it is already mentioned in "CP Ticketschalter" :)

therobrob:
The german pages are ready to release now; @lenderom translates them into english and french (thanks!)

PR #284, opened 2025-08-24
Fixed z-index in country map

therobrob:
## Description

The z-index of the country map was too high, so it was visible in front of the mobile menu. I fixed that.
IMG_8608

## Checklist

- [ ] English
- [ ] German
- [ ] French


MoritzWeber0 (approved):
Thanks!

PR #303, opened 2025-09-09
feat: List countries without FIP acceptance

### LINKED ISSUE #290: Country Map
Grey out the country out which don't have FIP (with dark grey)

https://www.fipguide.org/de/country/

MoritzWeber0:
List countries without FIP acceptance, show them on the map & add a glossary to the map.

Resolves #290


lenderom (dismissed):

MoritzWeber0:
> Thanks! I'll take another look at the code later. First, a question: Do San Marino, Andorra, Monaco, and Cyprus accept FIP or should we mark them as well? :)

San Marino, Andorra and Cyprus have no railway as far as I know, so I can add them to the list. Maybe it's worth to have another list of countries without railway at all? In Monaco, trains are operated by SNCF, not sure how the acceptance is.

therobrob (approved):

PR #309, opened 2025-09-11
feat: Add operators to navbar & merge dropdown implementations

### LINKED ISSUE #292: Rearrange the navigation bar
- Instead of "Choose Country" / "Länderauswahl", use "Countries" / "Länder"
- Add "Operators" as additional item to the navigation bar
- Move "News" and "Übergreifende Infos" to the left. Keep the dark mode toggle on the right.
- Do not underline the links in the navbar by default, only on hover (same applies to Contact, Imprint and Privacy Policy in the footer).
- Align the dropdown menu implementation with the language switcher implementation

MoritzWeber0:
Resolves #292


lenderom:
Could you check the mobile navigation? I think it looks a bit bugged. Maybe we don't need the border in the mobile navigation. :)

But the Desktop Version is amazing 🤩

MoritzWeber0:
> Could you check the mobile navigation? I think it looks a bit bugged. Maybe we don't need the border in the mobile navigation. :)
> 
> But the Desktop Version is amazing 🤩

I'm not 100% happy with the mobile navigation too, but without borders is also not optimal. Maybe we can try a few options and have a look at it together.

therobrob (approved):
i've identifies just two little aspects:
* When the mobile menu is open, only the elements within the menu should be focusable. We implemented this on deutschebahn.com, but unfortunately I can't remember exactly how we did it. :'D
* ESC-Support for mobile menu

We can outsource it into a new issue, if you like :)

PR #310, opened 2025-09-12
feat: Mobile content-navigation-menu

therobrob:
## Description
Display a content-navigation-menu on mobile devices (max. MD).

- sticky menu bar with page name
- bottom sheet: choose another chapter
- click and drag-support

### DIFF
--- a/assets/sass/contentNavigation.scss
+++ b/assets/sass/contentNavigation.scss
@@ -109,7 +109,7 @@
     border-radius: var(--border-radius-m) var(--border-radius-m) 0 0;
     transform: translateY(calc(100% - 6.4rem - env(safe-area-inset-bottom)));
     box-shadow: var(--box-shadow);
-    max-height: calc(100vh - 15rem - env(safe-area-inset-bottom));
+    max-height: calc(100vh - 20rem - env(safe-area-inset-bottom));
 
     &--open {
       transform: translateY(0);


therobrob:
Changes:
- Remove the "current chapter" text
- Instead, display the country/operator name
- Replace the button with a drag indicator

therobrob:
there's one problem left: `position: fixed` doesn't work with the dynamically changing viewport-height on mobile devices – the browser-command-line appears and disappears on scrolling. So it probably has to changed to `position: sticky`.

therobrob:
The mobile content navigation is ready for review! :)
Some annotations:
- i'm sure, there is some potential for code improvement – feel free to comment or adjust.
- the dragging-mechanism isn't perfect jet. The `deltaY` and the according `ifs` have to adjusted, but i wanted to finish this first implementation, because it took a long time.
- I've added an issue to unite the overlay implementations: https://github.com/fipguide/fipguide.github.io/issues/349

lenderom:
Great job @therobrob 🤩

I have two minor ideas :)
1. What do you think about adding the back to country option also above the headline for easier usage. So in this case: <- Back to Portugal 
Screenshot_20251004_103020_Chrome

2. What do you think about closing the menu when clicking on the greyed out content part of the page:
Screenshot_20251004_102934_Chrome

assets/sass/contentNavigation.scss
MoritzWeber0, line 1: 🎨 (non-blocking)

The border radius doesn't match in dark mode.

Image

assets/js/contentNavigation.js
MoritzWeber0, line 1: ❓ (question)

Is it intentional that only half of mobile navigation is displayed when clicking on the drawer? I have to click it another time to expand it completely.

Image

layouts/partials/contentNavigation.html
MoritzWeber0, line 1: 🐛 (bug)

The news pages are broken: 

Image

Example: https://deploy-preview-310--fipguide.netlify.app/en/news/6/

MoritzWeber0 (commented):
Thanks, the mobile navigation menu looks great! 🥳 

Before I start with the detailed review, here are some visual spots that I noticed.

assets/sass/contentNavigation.scss
MoritzWeber0, line 1: 🐛  (bug)

The mobile navigation on the PKP page is broken: 

Image

lenderom:
> > Great job @therobrob 🤩
> > I have two minor ideas :)
> > 
> > 1. What do you think about adding the back to country option also above the headline for easier usage. So in this case: <- Back to Portugal
> > 2. What do you think about closing the menu when clicking on the greyed out content part of the page:
> 
> Thanks :)
> 
> 1. you can go back to country in the open bottom-sheet already. Do you think, it is that important to add it at the top of the page? I wonder if it's too prominent there 🤔
> 2. This idea is included in [Unite overlay implementations #349](https://github.com/fipguide/fipguide.github.io/issues/349). Should we add it in this PR already?

1. This is also fine for me. It was just a quick idea that came to my mind :)
2. Ah, thanks haven't seen it :D

assets/sass/contentNavigation.scss
therobrob: in my dev mode in Edge browser it works properly, mhmm..
Bildschirmfoto 2025-10-06 um 22 15 02

therobrob: fixed in the newest commit :)

MoritzWeber0: Can you change the dimensions to iPhone SE? It seems to be a specific issue with these dimensions.

therobrob: i've added `flex-flow: column;` to `.o-aside__bottom-sheet`. This seems to work.

layouts/partials/contentNavigation.html
MoritzWeber0: It is fixed indeed, but now the content navigation is no longer there on mobile. Can we add it, similarly to the operator and country pages?

layouts/partials/contentNavigation.html
```diff
@@ -0,0 +1,87 @@
+<aside id="aside" class="o-aside o-aside__bottom-sheet">
```
MoritzWeber0, line 1: Since the same HTML is used for the aside and the bottom-sheet, shouldn't we just take `o-aside` and remove the "bottom-sheet" prefix?

`o-aside__bottom-sheet` > `o-aside`
`o-aside__bottom-sheet-header` > `o-aside__header`
`o-aside` > `o-aside__drag`
and so on.

The only conflict I see it the news sidebar, but maybe we can consistently update the classes there.

layouts/partials/contentNavigation.html
```diff
@@ -0,0 +1,87 @@
...
+  <button
+    class="o-aside__bottom-sheet-header"
+    aria-label="{{ T "contentNavigation.button" }}"
+    aria-controls="bottomSheet-content"
```
MoritzWeber0, line 5:
```suggestion
    aria-controls="bottom-sheet-content"
```

The id of the content element is different.

assets/sass/styles.scss
```diff
@@ -9,7 +9,7 @@ body {
   background-color: var(--bg-neutral);
   font-size: 1.8rem;
   overflow-x: hidden;
-  margin: 0;
+  margin: 0 0 7rem 0;
```
MoritzWeber0, line 12: The margin should only be applied if we are in mobile view and on a page where the mobile navigation is visible. For example not on the General Information page or on desktop devices.

assets/sass/contentNavigation.scss
```diff
@@ -0,0 +1,169 @@
...
+    top: unset;
+    bottom: 0;
+    left: 0;
+    width: 100%;
+    max-height: 100vh;
+    gap: 0;
+    overflow: hidden;
+    flex-flow: column;
+    transition: transform 0.5s ease-in-out;
+    border-radius: 1rem 1rem 0 0;
```
MoritzWeber0, line 109:
```suggestion
    border-radius: var(--border-radius-m) var(--border-radius-m) 0 0;
```

assets/sass/contentNavigation.scss
MoritzWeber0: Small detail, but there are left and right borders on the upper element, but not the lower element. Would be good to make this uniform. 

image

PR #316, opened 2025-09-25
feat: Increase gap between text and image on startpage

MoritzWeber0:
On one of my screens, the gap between text and image on the startpage is a bit small:


therobrob (approved):

MoritzWeber0:
> On which screen size does it occur? The headline should not span over a whole row. 🤔

2880x1800 with 90% scaling. The scaling is probably the issue.

PR #351, opened 2025-10-05
print view styling

therobrob:


therobrob:
Thanks @MoritzWeber0, the decreased font-size looks good for me :) 
If you're fine with the code, i would merge it with bypass rule, because i'm the initial author and you the last pusher :D

PR #355, opened 2025-10-06
fix: Do not translate icons

MoritzWeber0:
Do not translate the icon font as this leads to a broken display of icons:
image


therobrob (approved):
Cool, learned something new! Thanks for fixing it!

PR #367, opened 2025-10-07
fix: Add back missing margin for paragraphs in booking providers

MoritzWeber0:
In booking providers, each section has it's own paragraph. The CSS rule removed the margin for these paragraphs as it was the last element in relation to the parent.

This leads to missing margins, here in the example between the text and the Reservations heading:
image

After the change:
image

### DIFF
--- a/content/booking/bdz-ticket-office/index.de.md
+++ b/content/booking/bdz-ticket-office/index.de.md
@@ -21,5 +21,5 @@ Am BDŽ Ticketschalter können FIP 50 Fahrkarten für Reisen innerhalb Bulgarien
 {{% booking-section "reservations" %}}
 **Reservierungen**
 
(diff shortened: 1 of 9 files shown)


therobrob:
The css-rule is for the last paragraph in the article, because otherwise there is too much space and it doesn't look consistent. Maybe you could adjust the rule so that it doesn't affect the (booking) expanders? :)
Bildschirmfoto 2025-10-07 um 22 39 55

MoritzWeber0:
> The css-rule is for the last paragraph in the article, because otherwise there is too much space and it doesn't look consistent. Maybe you could adjust the rule so that it doesn't affect the (booking) expanders? :)
> Bildschirmfoto 2025-10-07 um 22 39 55
> 

Makes sense. I'll check how the rule could look like to achieve the behaviour.

MoritzWeber0:
> The css-rule is for the last paragraph in the article, because otherwise there is too much space and it doesn't look consistent. Maybe you could adjust the rule so that it doesn't affect the (booking) expanders? :)

I modified the rule, feel free to have another look.

PR #391, opened 2025-10-11
feat: Make search usable from all pages

therobrob:
- added search for all pages
- refactored search implementation
- refactored overlay implementation
- structure changes in i11n


therobrob:
thanks Moritz for reviewing and the styling update! i'm going to have a look!

PR #396, opened 2025-10-14
fix: Use relative position for toc indicator during scroll on mobile

MoritzWeber0:


therobrob:
i think the small gap is present in light _and_ dark mode. We have the border around the content-element, which is transparent in light mode. I've made it visible with this lightgreen color :)

Bildschirmfoto 2025-10-18 um 12 13 41

Bildschirmfoto 2025-10-18 um 12 13 28

MoritzWeber0:
> i think the small gap is present in light _and_ dark mode. We have the border around the content-element, which is transparent in light mode. I've made it visible with this lightgreen color :)

Thanks! I removed the borders on the left & right on mobile to avoid the issue.
Also removed the mixin, it was too complex for what it was doing and replaced it with two simple CSS rules.

PR #405, opened 2025-10-14
feat: Add train category expander

MoritzWeber0:
This PR introduces a new train category expander. The goal is a better consistency across pages and a better explanation of the icons used in the expander summary.

Part of the PR is also a refactoring of the text highlight styles to use the same logic for tags and text highlights.

To avoid too many merge conflicts, we should merge the technical implementation before adjusting all country pages as this might take some time. A helper has been added to the opencode commands and can by used by invoking `/update-traincategory {operator_name}`.

### DIFF
--- a/content/operator/sncf/index.de.md
+++ b/content/operator/sncf/index.de.md
@@ -36,84 +36,137 @@ Es gilt eine Reservierungspflicht in allen `TGV`, fast allen `IC` Zügen sowie e
 
 ### Fernverkehr
 
-{{% expander "Train à grande vitesse inOui (TGV inOui) ⚠️ℹ️" traincategory "long-distance" %}}
-**Beschreibung:**
-Der `TGV` inOui ist der Hochgeschwindigkeitszug von SNCF Voyageurs und verbindet zahlreiche Städte in Frankreich sowie internationale Ziele (z. B. München, Frankfurt am Main, Barcelona, Luxemburg, Brüssel, Zürich, Mailand). Eine Übersicht der Strecken ist [hier](https://www.sncf-connect.com/assets/media/2021-05/2014_axes-tgv_0.pdf) verfügbar. Im Wagen existieren alle Platznummern doppelt, der reservierte Platz ist der, an dem die Zahl beleuchtet ist. Der andere Platz hat auf der jeweiligen Fahrt keine Gültigkeit.
+{{% train-category
+  title="Train à grande vitesse inOui (TGV inOui)"
+  type="highspeed"
+  fip_accepted=true
+  reservation_required=true
+  route_overview_url="https://www.sncf-connect.com/assets/media/2021-05/2014_axes-tgv_0.pdf"
+%}}
+Der `TGV` inOui ist der Hochgeschwindigkeitszug von SNCF Voyageurs und verbindet zahlreiche Städte in Frankreich sowie internationale Ziele (z. B. München, Frankfurt am Main, Barcelona, Luxemburg, Brüssel, Zürich, Mailand). Im Wagen existieren alle Platznummern doppelt, der reservierte Platz ist der, an dem die Zahl beleuchtet ist. Der andere Platz hat auf der jeweiligen Fahrt keine Gültigkeit.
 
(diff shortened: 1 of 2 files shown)


layouts/partials/train-category.html
```diff
@@ -0,0 +1,131 @@
+<details class="o-expander o-expander--train-category">
```
therobrob: we should add the name-property so that only one expander is open in a time.
it could be managed by using `type`. what do you think? :)

assets/sass/_variables.scss
```diff
@@ -54,16 +76,27 @@ html[data-theme="dark"] {
...
+    }
+
     --pagefind-ui-border: #555;
     --box-shadow: 0 0.4rem 1rem rgba(0, 0, 0, 0.5);
     --box-shadow-light: 0.4rem 0.4rem 0.4rem rgba(0, 0, 0, 0.3);
     --pagefind-ui-background: var(--bg-default);
     --highlight-color-tip: #1a4a5c;
     --highlight-color-inofficial: #2a2d30;
     --highlight-color-important: #4a2a1a;
+    --highlight-color-confusion: #4a3a1a;
```
therobrob: is the little difference between `important` and `confusion` intended?

assets/sass/_variables.scss
```diff
@@ -54,16 +76,27 @@ html[data-theme="dark"] {
     --link-special: #ffffff;
     --bg-default: #151b23;
     --bg-neutral: #0d1117;
+    --bg-accent: #86761a;
+    --bg-accent2: #daba00;
+    --bg-accent3: #fff284;
```
therobrob: do we have these colors in light mode?

assets/sass/textHighlight.scss
```diff
@@ -54,3 +54,12 @@
...
 }
+
+.m-text-highlight--confusion {
+  background-color: var(--highlight-color-confusion);
+  border-left: #ff9100e1 solid 1rem;
+
+  @media print {
+    border: 0.2rem solid #ff9100e1;
+  }
+}
```
therobrob: isn't it the same color-schema like `highlight--important`? what's the need for adding the alpha here?

therobrob (commented):
Thanks a lot for your work on this 🫶 I've added some small comments / questions, could you check them? :)

layouts/partials/train-category.html
MoritzWeber0: The type isn't a strong enough differentiation in my opinion, we should rather just set one name for all train category expanders.

However, do we really want it in the end? Closing the other expander makes it harder to compare two train categories.

assets/sass/_variables.scss
MoritzWeber0: We don't need bg-accent2 and bg-accent3 anymore and I removed them from the light mode some time ago. Also removed them for the dark mode, good catch!

assets/sass/trainCategory.scss
```diff
@@ -0,0 +1,51 @@
...
+      > .m-text-highlight__roofline {
+        color: var(--tag-warning-color);
+      }
+
+      background-color: var(--tag-warning-bg);
+      border: var(--tag-warning-border) solid;
+      border-width: 1px 1px 1px 1rem;
+    }
+  }
+}
```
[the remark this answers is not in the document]
MoritzWeber0: The intention is to have consistent colors between the risk of confusion / important tag colors and the actual highlights. Users should be able to combine those by looking at the color. Since the implementation of highlight was different, I added a special case.

But the comment is valid and it adds additional complexity. So I checked some alternatives and ended up moving the highlight logic to match the tag logic. With the new logic, I can just reference the colors to ensure the same colors between highlight confusion, highlight important and the tag warning.

assets/sass/_variables.scss
```diff
@@ -14,6 +14,20 @@ $bg-code: #fff284;
...
+  error: #b70000,
+  info: #414141,
+);
+
+$tag-colors-dark: (
+  success: #a8d5ba,
+  warning: #f0d98d,
+  error: #f5a9ae,
+  info: #d1d1d1,
+);
```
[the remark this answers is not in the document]
MoritzWeber0: I removed the border mix, the contrast should be better now.

PR #409, opened 2025-10-15
fix: Apply border-radius to search scrollbar

MoritzWeber0:
Image

This PR fixes the missing border radius at the bottom right border.


therobrob:
without an border-radius it looks really unsexy :(
didn't know, some browser don't support it correctly, so i've added the border.

PR #411, opened 2025-10-15
feat: Dynamically resize search drawer on mobile

MoritzWeber0:
| Full screen | With keyboard visible |
|----------|----------|
| image | image |


therobrob:
Good Idea! Unfortunately it doesn't work on iOS Chrome and Safari :(
For the moment i've no idea how to fix this.
IMG_8961

PR #418, opened 2025-10-18
fix: Move dynamic search height to correct class

MoritzWeber0:
In #409, the definition of the max-height was moved to another class, this PR reflects this for the mobile max-height, otherwise it doesn't overwrite the value.

Also remove an unnecessary hard-coded positioning of the search drawer, which led to gaps between drawer and search bar on some devices.


therobrob (approved):
👍🏻

PR #420, opened 2025-10-18
fix: naming of booking options

therobrob:
On the booking page there was no difference between Trainseurope Ltd. phone and email

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -27,9 +27,9 @@ const initSearch = () => {
     },
   });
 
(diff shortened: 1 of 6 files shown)


therobrob:
> @therobrob I added a consistent naming. For example someone from abroad doesn't know that a "DB Reisezentrum" is. So I named this consistently Ticket Office. In the booking description I explained further: "DB Ticket Offices are called DB Reisezentrum". Does this match with your thoughts? :)

Yes, thank you! :)

lenderom (approved):

Issue #349, opened 2025-10-04
Unite overlay implementations


therobrob:
For now, we have three overlay implementations. We should unite them into one.
I prefer the one in baseof.html (`id="overlay"`). It could display with several z-indices for the needed usecase.

Overview of implementations:
- baseof.html Z. 10 `<div id="overlay" aria-hidden="true"></div>`
- home.html Z. 36 `<div class="curtain" aria-hidden="true"></div>`
- header.html Z. 13 `<div class="o-header__curtain"></div>`

In this context the following could added:
- resize-handling
- click on the overlay to close the open element (contentNavigation, Search, mobile Menu)

MoritzWeber0:
We could also add a small animation that slowly adds blurring.

PR #424, opened 2025-10-18
fix: Allow closing the search via the search button

MoritzWeber0:
Especially on mobile devices, space is limited and users may not have enough space to close the search if they want to. This PR changes the "open search" behavior to a "toggle search" behavior (open if search is closed, close if search is open).

In addition, two bugs are fixed:
- The overlay wasn't displayed when the user navigated to the search field on the home page via tab.
- The overlay wasn't closed if the search was opened via click and then moved out using the tab character (this will probably change when we implement the focus trapping, but for now it's the best solution I think).

### DIFF
--- a/content/operator/ns/index.de.md
+++ b/content/operator/ns/index.de.md
@@ -100,10 +100,6 @@ Regionalzüge für kurze Verbindungen mit Halt an jedem Bahnhof auf der Strecke.
 **Reservierungspflicht:** nein
 {{% /expander %}}
 
-{{% highlight important %}}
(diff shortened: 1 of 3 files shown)


assets/js/search.js
```diff
@@ -52,35 +70,41 @@ const initSearch = () => {
...
-    search.scrollIntoView({ behavior: "smooth", block: "start" });
+    openSearch();
   }
 
-  function showSearchOnStartPage() {
-    overlay.classList.add("overlay--show", "overlay--show-lv5");
-    searchElement.focus();
-    search.scrollIntoView({ behavior: "smooth", block: "start" });
+  function toggleSearchOnStartPage() {
+    if (overlay.classList.contains("overlay--show")) {
```
therobrob: In #349 we are going to unite the overlay implementations, so probably the open mobile menu also gets `overlay--show`. I think we will have an miss-behaviour of the of the overlays, if we unite them.

MoritzWeber0:
The scroll handler is broken again, I have to fix it.

PR #435, opened 2025-10-22
fix: Improve highlight color

MoritzWeber0:
Use a highlight color with better visibility, most relevant for the dark mode: 

Before:
image

After:
image


assets/sass/_variables.scss
```diff
@@ -51,9 +47,6 @@ html[data-theme="dark"] {
   --link-special: #ffffff;
   --bg-default: #151b23;
   --bg-neutral: #0d1117;
-  --bg-accent: #86761a;
-  --bg-accent2: #daba00;
-  --bg-accent3: #fff284;
```
therobrob: this variable is used for `<code>`-elements. These ones are without color now:

Image

PR #436, opened 2025-10-22
fix: Improve accuracy of toc highlighting

### LINKED ISSUE #414: Highlighting in toc still not accurate in all cases
Go to https://www.fipguide.org/en/operator/sncf/#special-tariff-conditions. "Special Tariff Conditions" should be highlighted, but instead "Reservation requirement in regional trains" is highlighted. Maybe that's because of https://github.com/fipguide/fipguide.github.io/pull/370

MoritzWeber0:
Resolves #414

The `.o-container` position has been changed to relative and there is no need to remove the scroll padding.

### DIFF
--- a/assets/js/search.js
+++ b/assets/js/search.js
@@ -38,50 +38,68 @@ const initSearch = () => {
 
   const searchElement = search.querySelector("input");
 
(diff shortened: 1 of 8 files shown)


therobrob (approved):
LGTM

PR #442, opened 2025-10-23
fix: Improve print version in dark mode

### LINKED ISSUE #365: Dark mode breaks print version
When dark mode is applied (not set to auto), the print version has the dark mode background:

Image

MoritzWeber0:
Resolves #365


therobrob (dismissed):
LGTM :)

PR #465, opened 2025-11-14
fix: bing webmaster improvements

therobrob:
- added data-decorative="true" to images, bc it should prevent bing to produce an error
- added missing meta descriptions
- lengthen the meta descriptions of the operator to 140 – 160 characters (contains the most important topics of the content)


MoritzWeber0 (approved):
Looks good, but I see some potential to automate the generation of descriptions. I'll work on it and open it as additional PR.

PR #468, opened 2025-11-16
feat: backlinks from booking page

therobrob:
- added index "booking" to all operator pages
- modified related-partial
- added backlinks from booking page to operator page

Because the index is on the operator pages, we need an different logic for related-partial than before.


content/operator/cd/index.de.md
```diff
@@ -5,6 +5,11 @@ description: "Informationen zu den FIP-Bedingungen bei der České dráhy (ČD).
 country:
   - "czechia"
 operator: "cd"
+booking:
+  - "cd-website"
+  - "db-website-fip-db"
+  - "db-website-fip-international"
+  - "cd-ticket-office"
```
MoritzWeber0: I'm not sure if tracking the booking platforms in another place is a good idea. I'm sure that the list will get out-of-sync at some point. Instead, we could automatically derive the ticket offices by iterating over the operators and check if the raw content contains `booking id="%s"`. It's less performant during build time, but given the limited number of operators it shouldn't be an issue. What do you think about the idea?

lenderom:
@therobrob Thanks for implementing this ❤️

PR #485, opened 2025-11-29
feat: Use different icons for booking expander meta

MoritzWeber0:
- Improve screen reader support, they can now read "FIP 50 Tickets bookable" or "FIP Global Fare not bookable", before it tried to read the icon as "no entry sign" or "check mark".
- Use color on the text for second identification of the value (color in addition to the icon).
- Get rid of the unicode icons and use Material Icons instead for better consistency.
- Text is now left-aligned.


therobrob:
I like the implementation! There was one small issue: in light mode the contrast was insufficient (min. 4,5:1 for small text). I've adjusted it :)

therobrob (approved):

PR #501, opened 2025-12-14
feat: operator Stena Line BV

therobrob:
added Stena Line BV as operator


therobrob:
The "General Information" page states that with Stena Line BV, one field on the free ticket is only valid for two crossings. What is the source for this information? We should mention this on the operator's website. @lenderom

PR #509, opened 2025-12-15
fix: Render footnotes in train category expander

MoritzWeber0:
Pass the train category expander markdown to the parent page to render it in the page scope, including support for footnotes. To make this possible, the scraping of new lines around the content has been removed. The small detail is described here in the [goldmark docs](https://github.com/yuin/goldmark/blob/master/_benchmark/go/_data.md):

> The only restrictions are that block-level HTML elements —
> e.g. `<div>`, `<table>`, `<pre>`, `<p>`, etc. — must be separated from
> surrounding content by blank lines, and the start and end tags of the
> block should not be indented with tabs or spaces.

### DIFF
--- a/layouts/partials/train-category.html
+++ b/layouts/partials/train-category.html
@@ -108,7 +108,7 @@
   </summary>
   <hr aria-hidden="true" />
   <div class="o-expander__content o-train-category__content">
-    {{- .content -}}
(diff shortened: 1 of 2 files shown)


therobrob (approved):
LGTM

PR #510, opened 2025-12-16
feat: Many train category expander improvements & SP migration

MoritzWeber0:
- Update archetypes to reflect the train category changes
- Avoid showing the Reservation heading within expanders in toc
- Add ship as train category type
- Add new "Scheduled routes" section within train category expanders
- Add id attribute for HTML anchors to create links to operators directly (expander is opened if referenced via link)
- Migrate SP & SBB

### DIFF
--- a/content/operator/sbb/index.de.md
+++ b/content/operator/sbb/index.de.md
@@ -65,7 +65,7 @@ Internationale Züge Richtung Deutschland und Italien.
 1\. Klasse: 13€ \
 2\. Klasse: 11€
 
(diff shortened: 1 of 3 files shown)


MoritzWeber0:
> Thank your for your migration efforts ❤️
> 
> What do you think about having informations about the reservations as an own section seperated from the text at the end? Like you have it done at SNCF? This way users can clearly see information about the reservation. I wouldn't do this a 100% strict rule but I think we should try to stick to it.
> 
> What do you think? :) image

I tried to to group the reservation information below the reservation headings. In addition, I added a new section called "Scheduled routes" for the overview of lines.

content/country/france/index.fr.md
lenderom: Also here

content/country/italy/index.fr.md
lenderom: More

lenderom: more more

content/country/spain/index.fr.md
lenderom: I'm running out of ideas

content/country/united-kingdom/index.fr.md
lenderom: Dum dum da dum

PR #512, opened 2025-12-19
feat: added aside in mobile view

### LINKED ISSUE #382: Mobile content navigation for news pages
We should decide if we want to have a mobile content navigation for the news pages and what content to show in it. 

So far, we show "Other news", "Related countries / operators" and the "Edit page" button in the desktop version, but not on mobile: 

Image

Since editing on mobile is hard, maybe we can show an alternative "Report an error / Suggest an improvement" button.

therobrob:
until now the booking aside was not displayed in mobile view. To link the booking information with the operators i’ve displayed the aside unterneath the content container.

it also closes #382 with fixing mobile content navigation margin.

### DIFF
--- a/assets/sass/styles.scss
+++ b/assets/sass/styles.scss
@@ -19,8 +19,9 @@ body {
   }
 }
 @media (max-width: #{$breakpoint-lg}) {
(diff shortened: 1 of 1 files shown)


MoritzWeber0 (approved):
Thanks 🫶

PR #514, opened 2025-12-19
feat: add multi-language 404

therobrob:
added multi-language 404 page

### DIFF
--- a/assets/sass/styles.scss
+++ b/assets/sass/styles.scss
@@ -388,6 +388,13 @@ details > summary {
   }
 }
 
+.o-link__mail,
+.o-link__tel {
+  & > .material-symbols-rounded {
+    margin-right: 0.4rem;
+  }
+}
+
 .overlay--show {
   backdrop-filter: blur(4px);
   background-color: rgba(0, 0, 0, 0.6);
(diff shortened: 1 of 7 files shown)


therobrob:
@MoritzWeber0 i’ve tried this one, but in local development and netlify preview i couldn’t check, if it works. Could you check it? :)

MoritzWeber0:
> @MoritzWeber0 i’ve tried this one, but in local development and netlify preview i couldn’t check, if it works. Could you check it? :)

The multi-language 404 page does work locally, but it will always open the English version of the page when deployed via GitHub. The reason is this line in the workflow that we have to update: https://github.com/fipguide/fipguide.github.io/blob/main/.github/workflows/deploy.yaml#L71.

There has to be a small 404 wrapper that redirects to the correct language. But it's already better than before, so I'd be fine with merging already.

MoritzWeber0 (approved):

MoritzWeber0:
> I’m not exactly sure what needs to be done, so it would be great if you could take care of it.

I implemented it, feel free to have a look around. I make use of the aliasing feature that we also use for the language detection on the home page, but rewrote it using modern methods (navigator.languages is commonly available in modern browsers).

PR #515, opened 2025-12-19
fix: updated available countries

therobrob:
I’ve missed to remove germany from content-not-available-yet.json correctly... 🙄


lenderom:
Ah damn I also just saw the removal of "availabla shortly" 😂

lenderom (approved):

PR #532, opened 2025-12-22
 feat: Add instructions to determine the Stena Line fare 

MoritzWeber0:
Adds instructions how to determine the FIP fare for Stena Line crossings.


therobrob (dismissed):
Cool that this works! Let's see how long the old booking system will last. 🙃

therobrob (approved):

PR #551, opened 2025-12-28
feat: performance optimizations

therobrob:
I’ve experimented with some performance optimizations. Feel free to have a critical view. :)

- merged pagefind implementations, because we already overwrite an large part of it. Copilot helped me to reorder the file.
- there was an issue regarding our fonts, which weren’t loaded with font-swap. So i’ve added it and Metric Overrides to avoid layout shifts
- the hero-image wasn’t loaded with `fetchpriority=high`
- there was an issue with chaining critital requests. I’ve added the defer property to the js.

### DIFF
--- a/layouts/partials/image.html
+++ b/layouts/partials/image.html
@@ -50,7 +50,7 @@
     data-decorative="true"
   {{- end }}
   loading="{{ $loading }}"
-  {{ with .fetchpriority }}fetchpriority="{{ . }}"{{ end }}
+  {{- with .fetchpriority -}}fetchpriority="{{ . }}"{{- end -}}
   srcset="{{ delimit $srcset ", " }}"
   sizes="(max-width: 576px) calc(100vw - 3.2rem), (max-width: 768px) 540px, (max-width: 992px) 720px, (max-width: 1200px) 960px, (max-width: 1400px) 1140px, 1320px"
   style="{{ delimit $styles "; " | safeCSS }}"


layouts/partials/head.html
```diff
@@ -19,7 +19,6 @@
     as="style"
   />
 {{ end }}
-<link rel="stylesheet" href="/pagefind/pagefind-ui.css" />
```
MoritzWeber0: The styling of the search seems to be broken, maybe that's because of some missing styles that were in the CSS before? 

Image

assets/sass/search.scss
```diff
@@ -9,142 +9,198 @@ $search-z-index: 12;
...
+    &__results {
+      padding-left: 0;
+      margin-bottom: 0;
+    }
 
-    &::placeholder {
-      opacity: 0.7;
+    // ---------------------------------------------
+    // SEARCH INPUT
+    // ---------------------------------------------
```
MoritzWeber0: Good comments should not repeat the code. Since the class is named search-input, the comment doesn't provide any additional value. I assume it was added by Copilot (and Copilot loves adding comments), so I've removed it and the other comments.

Nevertheless, here is a good read about best practices when writing comments: https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/#h2-b7593c698a940

assets/sass/search.scss
MoritzWeber0: On production, the title has a larger font size than the other text: 

Image

We should add it back.

Image

MoritzWeber0: The font of the search seems to be Arial, no longer Roboto in Chrome:


Image

Image

MoritzWeber0 (dismissed):

MoritzWeber0 (changes requested):
I only have one remark left: The loading animation seems to be broken after the changes (or is the change intentional?):

On the main branch:
Image

On this brach:
Image

PR #565, opened 2026-01-02
feat: Add missing Eurostar prices

### LINKED ISSUE #65: FInd out missing Eurostar prices
We should find out the missing Eurostar prices

Image

MoritzWeber0:
Add the missing Eurostar prices and remove the Brussels - Netherlands relation from Eurostar Blue as this doesn't seem to be bookable (it's either exit only or entry only): 

grafik

Resolves #65


lenderom (approved):
Thanks @MoritzWeber0 :)

PR #592, opened 2026-01-09
fix: FIP validity on Paris trams

MoritzWeber0:
FIP is no longer valid on the T12 and T13 lines as RATP took over the operation. Instead, FIP is now valid on the T9 and T14 lines.


MoritzWeber0:
I updated the information and it contains the latest of the available information now. T12 and T13 are no longer valid (RATP as operator is visible on the individual wikipedia pages of the lines, the wikipedia overview is not updated).

lenderom (approved):

Issue #481, opened 2025-11-24
[DESKTOP] Touch targets do not have sufficient size or spacing.


therobrob:
**Lighthouse Finding (Desktop) – macOS Edge, Browser**
Touch targets with sufficient size and spacing help users who may have difficulty targeting small controls to activate the targets. [Learn more about touch targets](https://dequeuniversity.com/rules/axe/4.10/target-size).

concerns the following:

- `a.o-aside__toc-link`

PR #596, opened 2026-01-10
fix: touch targets and aside margin in content navigation

### LINKED ISSUE #479: Margin-bottom inconsistently applied to content navigation
Image

The margin was introduced in https://github.com/fipguide/fipguide.github.io/pull/394. It doesn't work on nested lists. The margin should also be added between "Arrival and Border Points" and "Czech Republic" in the screenshot.

therobrob:
- increased size of footnote-backref and aside-links to meet wcag-requirements regarding touch target size
- added margin-top to nested li-elements in content navigation to achieve consistent space between the elements

Resolves #479 and #481.

### DIFF
--- a/assets/sass/contentNavigation.scss
+++ b/assets/sass/contentNavigation.scss
@@ -126,7 +126,7 @@
     gap: 0;
     overflow: hidden;
     flex-flow: column;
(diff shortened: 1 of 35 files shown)


MoritzWeber0 (approved):
Thanks ❤️

PR #598, opened 2026-01-10
fix: don’t render o-booking__classes if there is no content

### LINKED ISSUE #546: Booking expander renders always three columns
The booking expander is always rendered with three columns, even if there is no information
Example: https://www.fipguide.org/de/operator/stl/ (Stena Line Telefon)
Image

The text is wrapped, even if there is enough space.

therobrob:
don’t render o-booking__classes if there is no content

Resolves #546


MoritzWeber0:
The div is no longer rendered, so that's good, but the free space doesn't seem to be used by the `o-booking__meta` div, so the text still wraps.

grafik

therobrob:
> The div is no longer rendered, so that's good, but the free space doesn't seem to be used by the `o-booking__meta` div, so the text still wraps.

That’s absolutely intended, see https://github.com/fipguide/fipguide.github.io/issues/546#issuecomment-3732491247 :)

MoritzWeber0 (approved):

therobrob:
@MoritzWeber0 do you wanna merge #615 into this one before merging it? 🤔

PR #626, opened 2026-01-20
fix: vertically center aside links

MoritzWeber0:
They were not centered before:

image


therobrob:
> I changed if for the icons, but I'm not sure about the operators logos.
> 
> Which of the two options do you prefer? Left or right? @therobrob

Thank you! The operator-labels aren’t that long as the news-labels, so i prefer the right one. It looks better because text is centered vertically. This corresponds to the code of the PR, so it can be merged :)

PR #649, opened 2026-01-28
fix: dsb website international reservations

### LINKED ISSUE #648: DSB international seat reservations
The link "[DSB Travel B-Europe](https://travel.b-europe.com/dsb-rail/en/reservation-only)"of the booking page "[dsb website](https://www.fipguide.org/de/booking/dsb-website/)" is broken.
But there is the new website https://www.dsb.dk/en/international/, where the international reservations can be booked (for example Kopenhagen – Hamburg).

We should check, if we add this as an separate booking entry.

lenderom:
Closes #648

### DIFF
--- a/content/booking/dsb-international-website/index.en.md
+++ b/content/booking/dsb-international-website/index.en.md
@@ -5,8 +5,8 @@ description: "Booking information for the DSB International website."
 
 params:
   classes:
-    first: "30 DKK"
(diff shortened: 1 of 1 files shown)


therobrob:
#648 is about the new international dsb booking option. i think it should at least be mentioned, or better yet, listed as a separate booking option. What do you think? :)

PR #650, opened 2026-01-28
fix: add links to borderpoints

lenderom:
Adds internal links to the borderpoints table


therobrob:
What is the usecase for this change? :) do i look at the borderpoint and then wanna go to the corresponding country or operator? Why don’t we link the sections under the table either?

PR #668, opened 2026-02-01
refactor: Remove legacy traincategory expanders

MoritzWeber0:
Since the migration seems to be done after merging #667, #663, #660 and #659, we can remove the migration script and the old category in the expander implementation.


therobrob:
> > was it intended to delete the whole `.opencode/command/update-traincategory.md`-file?
> 
> Yes, the only intended purpose was the migration. Or do you see any other use case? :)

ah yes, sorry i didn’t get it :D

Issue #558, opened 2025-12-29
Add feedback to website


MoritzWeber0:
When we use feedback received via email as source on a content page, we create a GitHub issue with a screenshot of the email and reference it.

For platform independence and more flexibility, we could add those emails directly as pages with some styling. They should open as dialog / popup when clicking on the source.

Issue #446, opened 2025-10-24
Number of FIP Coupons and Reduction


lenderom:
Different operators have different aggreements how many FIP Coupons are issued for each operator, if relatives can also receive Coupons and the discount for FIP 50 Tickets. We should display these different information.

PR #680, opened 2026-02-08
feat: Add support for dialogs

MoritzWeber0:
Dialogs are a helpful element for presenting content without cluttering the main page. They are basis for the implementation of #446 and #558.

A dialog usually consists of two parts: The trigger and the dialog content.

If you want to use dialogs in the Markdown content, we provide two helpers for dialog creation and linking.
To add the dialog content, add the following shortcode to the respective page content:

```md
{{% dialog
  id="test"
  title="Test"
%}}
This is the content of the dialog.
{{% /dialog %}}
```
**IMPORTANT:** The dialog id must be unique on the relevant page.

As you'll notice, there is no change to the page. We have to create a trigger. The easiest way is link-based trigger. You can reference the dialog by ID:

```
[This is the trigger for the dialog](dialog:test)
```

For better visualization, I added the "[FIP Guide Team email inquiry to Stena Line](https://deploy-preview-680--fipguide.netlify.app/de/operator/stl/#quellen)" source as dialog.

TODO:

- [x] z-index of overlay has to be increased to cover the header
- [x] For better contrast add border to dialog
- [x] New lines for email header
- [x] Button alignment on mobile


layouts/partials/dialog.html
```diff
@@ -0,0 +1,15 @@
+<dialog
+  id="{{ .ID }}"
+  class="o-dialog__wrapper"
+  aria-labelledby="{{ .ID }}-title"
+>
+  <div class="o-dialog o-container">
+    <div class="o-dialog__header">
```
therobrob:
```suggestion
    <header class="o-dialog__header">
```

layouts/partials/button-internal.html
```diff
@@ -0,0 +1,3 @@
+<button class="a-button a-button__internal">
```
therobrob: the "close dialog"-button must have a title-attribute like "Close" for a11y-reasons.

assets/sass/dialog.scss
```diff
@@ -0,0 +1,48 @@
...
+    border: var(--border);
+    overflow: hidden;
+  }
+
+  &__header {
+    display: flex;
+    column-gap: 1rem;
+    align-items: center;
+    justify-content: space-between;
+    padding: 1.6rem 2rem;
```
therobrob: For my taste, the padding-top/-bottom is a bit too large and i’d prefer 1rem.

assets/js/dialog.js
```diff
@@ -0,0 +1,60 @@
...
+    closeButton.addEventListener("click", () => closeDialog(dialog));
+  }
+}
+
+function closeDialog(dialog) {
+  dialog.close();
+  closeOverlay();
+}
+
+function closeAllDialogs() {
```
therobrob: could there be more than one open dialog? do you think of a dialog in a dialog? 😅

assets/js/dialog.js
```diff
@@ -0,0 +1,60 @@
...
+  dialog.show();
+  openOverlay("dialog");
+
+  const closeButton = dialog.querySelector(".o-dialog__header > .a-button");
+  if (closeButton) {
+    closeButton.addEventListener("click", () => closeDialog(dialog));
+  }
+}
+
+function closeDialog(dialog) {
```
therobrob: do we want to remove the event-listener from the closeButton while closing the dialog?

therobrob (commented):
thanks a lot for implementing <3

i’ve added some thoughts and ideas to the code. Besides this, I noticed the following. Feel free to look through it, I look forward to hearing your thoughts! :) Let’s talk, if something is unclear. :)

- after closing the dialog, the dialog-heading remains in the a11y-tree. maybe this is a bug of edge, but i didn’t want this unmentioned.
- there is no focus-trap in the opened dialog (could be fixed later on)
- on resize the overlay disappears. this should be an side effect of the overall overlay implementation.
- the dialog-content has to be copied into every .md-file, although it is only in one specific language (could be resolved later on)
- i like there‘s an icon next to the dialog trigger. so the user isn’t surprised of the dialog. i’ve experimented of some adjustments: 
1) we could change the order, so the icon comes first an then the text. For now the dialog-icon gets in the way of the reference icons slightly.
2) i’m convinced you played around with the google icons :D maybe these ones fit too: "pinch", "quick_reference_all", "read_more"

lenderom:
Im not sure if this is only a problem in #705 but it seems like the search doesn't work correctly with the dialogs. The information are found, but the dialog does not open automatically.

MoritzWeber0:
> > Im not sure if this is only a problem in #705 but it seems like the search doesn't work correctly with the dialogs. The information are found, but the dialog does not open automatically.
> 
> Oh yeah, we can implement something similar to the expanders for this 👍

Done, just need the approval from someone :)

therobrob (approved):

PR #684, opened 2026-02-08
feat: add issuer specific fip validity to operator pages

### LINKED ISSUE #446: Number of FIP Coupons and Reduction

MoritzWeber0:
For each FIP issues a file called `validity.yaml` can be added to the corresponding operator directory, including information which how many coupons for which operator are issued. For example, let's have a look the at `validity.yaml` for FIP issuer DB:

```yml
_anchors:
  coupon-4fields: &coupon-4fields
    status: valid
    text:
      de: "1 Freifahrtschein mit jeweils 4 Feldern pro Jahr. Jedes Feld ist zwei Tage gültig."
      en: "1 coupon with 4 fields each per year. Each field is valid for two days."
      fr: "1 coupon avec 4 champs chacun par an. Chaque champ est valable deux jours."
  coupon-not-available: &coupon-not-available
    status: invalid
    text:
      de: "Nicht verfügbar"
      en: "Not available"
      fr: "Non disponible"
  reduced-50: &reduced-50
    status: valid
    text:
      de: "50 % Rabatt"
      en: "50 % discount"
      fr: "50 % de réduction"
pkp:
  fip-coupon: *coupon-4fields
  fip-coupon-relatives: *coupon-4fields
  fip-reduced-ticket: *reduced-50
zssk:
  fip-coupon: *coupon-4fields
  fip-coupon-relatives: *coupon-not-available
  fip-reduced-ticket: *reduced-50
```

Resolves #446

TODO:

- [x] Add to archetype
- [x] Transform all shortcodes
- [x] Add Params from RDG
- [x] Add Params from Discord (https://discord.com/channels/1250522473188032512/1451295200743002163/1451295200743002163)
- [x] Obtain information from: https://discord.com/channels/1250522473188032512/1433782574806728804/1470447474345775198
- [x] Replace `__info-button` with internal button
- [x] Handle Euskotren FIP Ticket
- [x] Stena line two fields
- [x] Remove own operator from table
- [x] Unify DB text
- [x] Merge relatives and active employee coupon section in one table and separate retirees

### DIFF
--- a/assets/js/fipValidityComparison.js
+++ b/assets/js/fipValidityComparison.js
@@ -1,4 +1,4 @@
-const { closeDropdown } = require("./dropdown");
+import { closeDropdown } from "./dropdown.js";
 
 const ISSUER_KEY = "fipguide-issuer";
 
--- a/i18n/de.yaml
+++ b/i18n/de.yaml
@@ -55,7 +55,7 @@ fipValidity:
     Wähle zunächst den Aussteller deines FIP Ausweises. Wenn du berechtigt bist,
     dann kannst du die Freifahrtscheine über deinen Arbeitgeber beantragen.
     Bitte beachte, dass wir bei der Ausgabe der Freifahrtscheine nicht
-    weiterhelfen können. Einen Überblick aller Freifahrtscheine erhälst du
+    weiterhelfen können. Einen Überblick aller Freifahrtscheine erhältst du
     [hier](/fip-validity).
   fip-global-fare: FIP Globalpreis
   fip-reduced-ticket: FIP Ermäßigtes Ticket (FIP 50 / FIP 75)
--- a/layouts/partials/fip-validity/issuer-dropdown.html
+++ b/layouts/partials/fip-validity/issuer-dropdown.html
@@ -14,8 +14,9 @@
     <span
       data-fip-validity-label
       data-fip-validity-placeholder="{{ T "fipValidity.selectIssuerPlaceholder" }}"
-      >{{ T "fipValidity.selectIssuerPlaceholder" }}</span
     >
+      {{ T "fipValidity.selectIssuerPlaceholder" }}
+    </span>
     <div class="o-dropdown__icon">
       {{ partial "icon" "keyboard_arrow_down" }}
     </div>


MoritzWeber0:
> @MoritzWeber0 Thank you for your implementation. I think this will help many of our collegues. Do you think we should implement a kind of "Default" per employer? Like usually SNCF gives 1 Coupon with 4 Fields and we only specify the deviations on the operator site?

I'm against an implicitly added default. We should only add explicit information if we have it, otherwise let's fall back to unknown. The only reason would be less boilerplate (the YAML section grows a bit too quickly maybe). But there are probably better solutions like defining some defaults globally and then just reference them via a one-liner.

lenderom:
> > @MoritzWeber0 Thank you for your implementation. I think this will help many of our collegues. Do you think we should implement a kind of "Default" per employer? Like usually SNCF gives 1 Coupon with 4 Fields and we only specify the deviations on the operator site?
> 
> I'm against an implicitly added default. We should only add explicit information if we have it, otherwise let's fall back to unknown. The only reason would be less boilerplate (the YAML section grows a bit too quickly maybe). But there are probably better solutions like defining some defaults globally and then just reference them via a one-liner.

Yeah I was more thinking about the boilerplate, not a "default" information but a way to avoid writing on every page "SBB employees have 8 fields every field is valid for one day" :D
But it's not a must have. Just an idea to avoid extremly large yaml metadata :)

therobrob (commented):
I’ve just reviewed the technical parts, not the md-files.

Two further comments:
- for me it seems slightly confusing showing the grey status with question marks for cases, where we don’t know the information from all operators. It suggests we have no information _at all_. I think this case is very common, until we don’t have the information from all operators.
Edit: why do we handle the same case ("we don’t have information from all operators") in two different ways?
One time we say "valid" and one time we say "unknown" – where’s the difference between them?
Bildschirmfoto 2026-02-09 um 19 32 08

- this leads me to the second aspect: the button for the dialog isn’t optimal in my opinion. It is not self-explanatory, because there is no label and there is the second status-icon to the left (two icons, two different meanings for interacting). I envision a single, comprehensive button, so the user can click the whole information.

MoritzWeber0:
> Yeah I was more thinking about the boilerplate, not a "default" information but a way to avoid writing on every page "SBB employees have 8 fields every field is valid for one day" :D But it's not a must have. Just an idea to avoid extremly large yaml metadata :)

I changed the implementation to use a `validity.yaml` per FIP issuer. This has a few advantages:
- We can use YAML anchors to reuse texts, this reduces the amount of duplicated content, making it easier to maintain it in the future.
- The information is now reversed in comparison to the previous implementation. We collect the information per FIP issues in one a file. This represents how we get the information.
- We don't have to define the validity multiple times for the different languages, only the text is still localized.
- The frontmatter remains unchanged and we don't have to scroll endlessly to get to the operators content.

MoritzWeber0:
The table is now also available in the reverse format to find out for which operator how many coupons can be requested: https://deploy-preview-684--fipguide.netlify.app/de/fip-validity/

PR #688, opened 2026-02-09
fix: update NS information about ICD and ECD trains

lenderom:
Update the information for the NS ICD and ECD


lenderom:
https://deploy-preview-688--fipguide.netlify.app/de/operator/ns/

therobrob:
There is one issue left: there is no space between the images and the button.
Bildschirmfoto 2026-02-24 um 23 49 16
Bildschirmfoto 2026-02-24 um 23 49 27

We have to care about the other buttons. `o-booking__links` has an margin-top, but it would add an extra space here..

PR #740, opened 2026-03-07
feat: Add support for lightboxes

### LINKED ISSUE #739: Add lightbox for internal images
When clicking on an internal image, a dialog should open with the full version of the image. Dragging / Zooming should be possible.

Originally tracked in #314, but the issue is closed.

MoritzWeber0:
Add support for lightboxes to view images in a full-screen mode.

There are three different ways to open a lightbox:

1. Via markdown image. To enable lightbox mode for images, just pass an alt text. Images with alt are considered decorative and therefore don't have a lightbox.
   ```
   !["Image caption"]("/path/to/image.webp" "Here is the alt text")
    ```
1. Via float-image. Same as via markdown image, but the alt text is passed as parameter.
    ```
    {{% float-image
        src="image.webp"
        alt="Here is the alt text"
        caption="Image caption"
        width="50%"
        position="right"
    %}}
     ```
1. Via markdown link. Just pass the reference to the image like you would do with markdown images:
   ```
   ["Link text"]("/path/to/image.webp")
    ```

All of this is done without additional library, just using what we have. A combination of pan & zoom from the country map and the new dialog.

There are some additional changes related to pan zoom (I just moved the parts to separate files to make it independent from the map)


assets/sass/lightbox.scss
```diff
@@ -0,0 +1,24 @@
...
+  cursor: zoom-in;
+}
+
+.o-dialog__wrapper--lightbox {
+  .o-dialog__wrapper {
+    max-width: 90dvh;
+  }
+
+  .o-dialog {
+    max-height: unset;
```
therobrob: some images aren’t appear in full size
see https://deploy-preview-740--fipguide.netlify.app/de/operator/sll/ "Kosten online herausfinden" -> "Step 1"

Image

therobrob (changes requested):
Thanks Moritz :)
The following things came to my mind:
- do we really need zoom controls for images? i think it’s really uncommon, isnt’t it?
- the focus indicators should be set correctly
- is there a way to show images in the lightbox in full size? Some users might overlook the zoom or the option to move the image.

PR #784, opened 2026-03-21
feat: Add support for FIP75 in various places

MoritzWeber0:
- Add FIP 75 as option to booking platforms. Some booking platforms can book FIP 50 tickets, but no FIP 75 tickets (for example the DB website).
- Make the fip-validity subtitle customizable. It no longer mentions FIP 75 on operator pages where 75% discount is not available.
- Remove the discount section from GB and cover it with the new fip-validity. Update FIP validity to include the discounts that were represented as text on the GB page before.
- In the texts of SNCB, SNCF and GB mention FIP 50 and FIP 75 where it makes sense.
- Remove distinction between safeguarded and non-safeguarded for reduced rate tickets, the distinction only exists for FIP coupons.


content/operator/cie/index.de.md
therobrob: should we add the subtitle here?

PR #810, opened 2026-04-05
feat: redesign 404 page with error image and fix GitHub Pages routing

### LINKED ISSUE #136: Illustration for 404-Page
We could add an cool illustration to 404-Page

For example:
- A derailed train
- A train with a disruption
- A station sign with a delay/cancellation
- A train stopped at a red signal
- A switch with an incorrectly set signal

Relates to https://github.com/fipguide/fipguide.github.io/issues/67

### LINKED ISSUE #717: 404 Page not working
Check this for example: https://www.fipguide.org/betreiber

lenderom:
## Summary

- Redesign the 404 error page with a centered error illustration (`error.webp`) and improved layout using `float-image` and `intersector` shortcodes across all three languages (en, de, fr)
- Add homepage links to the "go to homepage" bullet point in all languages
- Fix GitHub Pages 404 routing: replace the Hugo alias-based `404.html` (which incorrectly landed under `/en/404.html` due to `defaultContentLanguageInSubdir: true`) with a `static/404.html` that reliably serves from the site root
- Add `center` position support for `float-image` in SCSS

Fixes #717
Fixes #136


assets/sass/floatImage.scss
```diff
@@ -11,4 +11,10 @@
     float: left;
     margin-right: 2rem;
   }
+
+  &--center > figure {
+    float: none;
```
therobrob: So the class .m-float-image—center doesn’t contain any floating? Is this the correct use of BEM?

content/404/index.de.md
```diff
@@ -4,12 +4,20 @@ params:
   pagefind_ignore: true
 ---
 
-Hoppla! Die gesuchte Seite ist nicht verfügbar oder wurde verschoben.
+{{% float-image
```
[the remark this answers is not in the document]
lenderom: Hmm as far as I know there is no way to change the size of an image?
I just wanted to have a centered image which is not 100% size :D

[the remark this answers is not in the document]
MoritzWeber0: What do you think of a split view instead? On the left site the image, on the right side the text? Right now, some parts of the text are centered, but the last part is left-aligned.

Image

static/404.html
```diff
@@ -0,0 +1,42 @@
...
+  <head>
+    <title>404 – Page not found</title>
+    <meta name="robots" content="noindex" />
+    <meta charset="utf-8" />
+    <noscript>
+      <meta http-equiv="refresh" content="0; url=/en/404/" />
+    </noscript>
+    <script>
+      (() => {
+        const supportedLanguages = ["en", "de", "fr"];
```
MoritzWeber0: Let's not hard-code the languages and iterative over the site languages instead.

static/404.html
MoritzWeber0, line 1: I'm a bit confused this is actually needed, in theory the alias should handle it exactly the same way. It also has a section for 404. There seems to be a regression in hugo that it stopped working, the `404.html` is no longer created automatically.

content/404/index.de.md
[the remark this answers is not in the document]
lenderom: On Desktop this could be a good solution. But I guess we would still have the same issue on mobile...

static/404.html
MoritzWeber0: The regression was in https://github.com/gohugoio/hugo/releases/v0.155.0, where support for multi-language aliases was added.

MoritzWeber0: It was marked as breaking change, but I overlooked it: https://discourse.gohugo.io/t/alias-paths-in-v0-155-0-and-later/56674

content/404/error.webp
MoritzWeber0, line 1: The contrast of the image in dark mode is pretty bad.

static/404.html
MoritzWeber0: So looks like the old approach via the alias is no longer supported. We should get rid of the 404 section in the alias then.

MoritzWeber0: Ok, the handling was just a bit different. This would be a hotfix for it, without too many changes: #831. What do you think of this approach?

lenderom: If this works, we can also do this :)

content/404/error.webp
lenderom: Feel free to change the image :)

static/404.html
MoritzWeber0: In the netlify preview it does, so it should also work in GitHub Pages :)

content/404/error.webp
lenderom: What do you think of the new one (source https://www.pexels.com/de-de/foto/30503335/)

layouts/404/single.html
```diff
@@ -0,0 +1,23 @@
...
+          (dict
+          "image" $image
+          "width" "100%"
+          "Page" .
+          )
+        -}}
+      </div>
+      <div class="o-error__divider">
+        {{ partial "intersector" }}
+      </div>
```
therobrob: I’d not show it overall, because it doesn’t divide two paragraphs.

layouts/404/single.html
```diff
@@ -0,0 +1,23 @@
+{{ define "main" }}
+  <article class="o-error" data-pagefind-ignore="all">
```
therobrob, line 2: I’d prefer this one:
`<article class="o-single o-single--error" data-pagefind-ignore="all">`

static/404.html
MoritzWeber0, line 1: We can get rid of the file now, right?

content/404/index.fr.md
```diff
@@ -4,13 +4,13 @@ params:
...
+Oups ! La page que vous recherchez est indisponible ou a été déplacée.
 
-Voici ce que vous pouvez faire ensuite :
+Voici ce que vous pouvez faire ensuite :
 
-- **Vérifiez l’URL :** Assurez-vous qu’il n’y a pas de faute de frappe dans l’adresse.
-- **Accédez à la page d’accueil :** Retournez à la page d’accueil et naviguez à partir de là.
-- **Utilisez la navigation :** Essayez de naviguer via le menu ou d’utiliser la fonction de recherche.
-- **Contactez-nous :** Si vous avez besoin d’aide, n’hésitez pas à nous contacter : [nextstop@fipguide.org](mailto:nextstop@fipguide.org)
+- **Vérifiez l'URL :** Assurez-vous qu'il n'y a pas de faute de frappe dans l'adresse.
```
MoritzWeber0:
```suggestion
- **Vérifiez l’URL :** Assurez-vous qu'il n'y a pas de faute de frappe dans l'adresse.
```

You probably used a Anthropic model. They have the bug that they replace `’` with `'`. It's a bit annoying. Can you change it back for all occurrences in the file?

layouts/404/single.html
MoritzWeber0: Not sure about it, the error case is pretty specific and isolated, we won't use it in another file. So I'd keep it in a separate SCSS file and not mix it in the content SCSS file. Another idea is to rename it to `o-404` or something like it.

therobrob: I didn’t talked about scss at all. The syntax and the file-system are two different things in my opinion. In BEM this is a special case for `o-single`, so we should treat it like that :)

MoritzWeber0: I always prefer a clear logic to find out in which SCSS file a CSS rule is stored and the prefix is the easiest one to remember. But we don't do that consistently, so yeah, we can also call it `o-single--error`.
However, we should then also update the other elements as well, for example `o-error__body--error` and so on.

content/404/index.de.md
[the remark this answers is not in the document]
therobrob: with the new image this is resolved. :)

content/404/error.webp
therobrob: LIKE! :)

layouts/404/single.html
therobrob: if we do so, we should update the other elements as well, yes 👍🏼  I agree with you to have a clear logic – let’s discuss the scss file structure in another issue :) https://github.com/fipguide/fipguide.github.io/issues/879

MoritzWeber0: Sounds good 👍

lenderom: I will close this conversation for now and keep the implementation as it is. Adressing this issue in #879

lenderom:
@MoritzWeber0 @therobrob can we merge this now? :)

therobrob (approved):

PR #827, opened 2026-04-10
fix: Let dropdown overflow the content in fip validity

MoritzWeber0:
Make this overflow possible: 

image


therobrob:
Did you intend for it to extend beyond the footer? I‘d rather like an dropdown with fixed height and `overflow-y: scroll` to prevent it to extend beyond the footer. The native browser/OS-drodowns neither behave this way. 🤔

MoritzWeber0:
> Did you intend for it to extend beyond the footer? I‘d rather like an dropdown with fixed height and `overflow-y: scroll` to prevent it to extend beyond the footer. The native browser/OS-drodowns neither behave this way. 🤔

That's a good idea I haven't thought of yet. I'll try it out when I have some for it 😊

MoritzWeber0:
Another issue with this implemention that needs fixing

MoritzWeber0:
The max-height is implemented in #1014, so this is no longer needed.

PR #829, opened 2026-04-10
fix: Avoid page jumps when search loads

MoritzWeber0:
Assign a fixed height to the search so that the space is already reserved before the search loads.


therobrob (approved):
LGTM

PR #837, opened 2026-04-11
 feat: Switch to pagefind component UI 

### LINKED ISSUE #538: Optimize search
When searching for "Schweiz" in the German start page, the operator pages are ranked higher than the actual Switzerland page. In the optimal case, the title does always match first.

MoritzWeber0:
In [pagefind v1.5.0](https://github.com/Pagefind/pagefind/releases/tag/v1.5.0), a completely new component UI library was added. While the old default UI remains supported, the recommendation is to switch to the new UI. It provides better accessibility, customization and has a modern touch. This is the implementation for the FIP Guide. It makes many things simpler, so I could remove a lot of custom code that was needed before for the same behavior.

It does also resolve #538: LKA is now listed as result and Schweiz is ranked as first result in the German search.


layouts/partials/head.html
```diff
@@ -9,6 +9,7 @@
   {{ end }}
 </title>
 {{ partial "head/js" (dict "file" "js/darkmode.js" "defer" false) }}
+{{ partial "head/search" . }}
```
therobrob, line 12: it is referenced in `<head>` but it appears in `<body>` and therefore the following elements (e. g. stylesheet, meta-information, ...) also appear in `<body>` – is this intended?

layouts/_default/home.html
```diff
@@ -12,12 +12,7 @@
       </div>
 
       <div class="o-startpage__search">
-        <div
-          id="search"
-          class="o-search o-search--startpage"
-          data-placeholder="{{ T "search.placeholder" }}"
-          data-label="{{ T "search.label" }}"
-        ></div>
+        <pagefind-modal-trigger></pagefind-modal-trigger>
```
therobrob: the screenreader lose the context information. Now the information is only "suchen", i think it should be "Suchbereich öffnen" or sth like that

assets/sass/search.scss
```diff
@@ -1,240 +1,96 @@
...
+  );
+  background-size: 100% 100%;
+  background-repeat: no-repeat;
+  border-radius: 0.5rem;
+  padding: 0.4rem;
+  margin: 0 -0.4rem;
+  animation: pagefind-highlight-reveal 260ms ease-out;
+
+  @media (prefers-reduced-motion: reduce) {
+    animation: none;
```
therobrob, line 77: the animation is funny ^^ the highlighted text is somewhat difficult to see because the coloring isn't very strong.

assets/sass/search.scss
therobrob, line 1: could we add the hover effect from the teaser cards to `.pf-result-card` and `.pf-heading-chip`? What do you think? :)

therobrob, line 1: It would be nice to add an outline-offset of 1px to the focus indicators :)

therobrob (commented):
This pagefind-update is really cool! So we're getting rid of some custom code :)
I’ve left some thoughts in the comments. Further more:
* the aria-label of the input element isn’t fitting: "Suche diese Seite" => "durchsuche diese Seite. We should [overwrite it](https://pagefind.app/docs/custom-components/#accessing-translations), what do you think?
* the screenreader read out the # from the .pf-heading-chip elements (for example "Nummernzeichen Wissenswertes“). Could we hide it fore screenreaders?
* does the pagefind version support an aria-live feature? I think we should announce the search results after typing in the search term.

assets/sass/search.scss
MoritzWeber0: Sure! Currently I already have it when focused, should be easy to extend it to the hover effect.

layouts/_default/home.html
MoritzWeber0: Sounds like a bug with pagefind itself. I found some more bugs, maybe I'll open a PR on their side, the other project could benefit then as well.

MoritzWeber0: It's possible to customize the text via the translation or by passing a placeholder, but then it also updates the placeholder text. So for now, I'll update it in the JS.

MoritzWeber0:
> This pagefind-update is really cool! So we're getting rid of some custom code :) I’ve left some thoughts in the comments. Further more:
> 
> * the aria-label of the input element isn’t fitting: "Suche diese Seite" => "durchsuche diese Seite. We should [overwrite it](https://pagefind.app/docs/custom-components/#accessing-translations), what do you think?

There is already an open PR for it, I would just wait until it's merged: https://github.com/Pagefind/pagefind/pull/1096

> * the screenreader read out the # from the .pf-heading-chip elements (for example "Nummernzeichen Wissenswertes“). Could we hide it fore screenreaders?

There was a way by overwriting the content of the `::before` element, so it should be hidden now.

> * does the pagefind version support an aria-live feature? I think we should announce the search results after typing in the search term.

Looks like they don't want to implement it, at least that's the only comment I could find regarding to it: https://github.com/Pagefind/pagefind/pull/162#issuecomment-1338269492

assets/sass/search.scss
MoritzWeber0: Yeah, it's a dilemma between visibility and contrast. Stronger highlight means less contrast for the text to the background. However, I made it a bit stronger now and added a separate rule for dark mode.

layouts/partials/head.html
MoritzWeber0: I moved the non-script parts out and added it them to the body in baseof.

assets/sass/search.scss
MoritzWeber0: Done. I removed the box-shadow as it was conflicting with the outline.

MoritzWeber0:
The comments should be addressed now :)

assets/js/search.js
```diff
@@ -1,127 +1,34 @@
...
-    element: "#search",
-    highlightParam: "highlight",
-    showSubResults: true,
-    translations: {
-      placeholder: placeholderText,
-      search_label: searchLabelText,
-    },
+function initSearchAriaLabel() {
+  document.querySelectorAll(".pf-trigger-btn").forEach((btn) => {
+    btn.setAttribute("aria-label", "Suchbereich öffnen");
```
therobrob: should we localize this?

assets/sass/search.scss
therobrob: thank you! can we add `transition: background .2s,color .2s` to the elements?

therobrob (commented):
There are two little this left, i’d like to address.

Do you think the localization-issue on behalf of pagefind is solved soon? Otherwise we can merge this PR before and update pagefind later on :)

Regarding aria-live: my idea is to announce, that the search results changed, not what exactly changed in the search results. I understood the pagefind-MR to mean that aria live should announce all changed results. But we can leave this unaddressed for now, i think there is no serious violation. :)

MoritzWeber0:
> Do you think the localization-issue on behalf of pagefind is solved soon? Otherwise we can merge this PR before and update pagefind later on :)

They are usually fast in merging PRs, so I would do it the way proposed: Merge this PR and then update pagefind once the new version is released.

> Regarding aria-live: my idea is to announce, that the search results changed, not what exactly changed in the search results. I understood the pagefind-MR to mean that aria live should announce all changed results. But we can leave this unaddressed for now, i think there is no serious violation. :)

Makes sense to me, but sounds like a bit of effort, so feel free to track it as issue if you want :)

assets/js/search.js
MoritzWeber0: Good catch, missed that. Done.

MoritzWeber0:
Done, I also replaced the border with outline (like we do it for the teasers), otherwise there were cases with two borders / outlines.

EDIT: Now the border is gone, I'll fix it tomorrow morning.

therobrob (approved):
Everything looks great now, thanks a lot! 🫶

PR #856, opened 2026-04-18
refactor: Render booking pages in operator context

### LINKED ISSUE #462: Check headings map on operator pages
Check if the headings in booking expanders are properly recognized as part of the expander in the headings map.

### LINKED ISSUE #852: Remove single booking pages as direct links
They should still remain on the operator pages, but the separate pages should be removed from the index.

### LINKED ISSUE #887: Handling of expanders in search
Steps to reproduce:

1. Search for "Interregio" in the German version of the page
1. Select the "Reservation" result: 
    Image
1. The Intercity expander opens instead of the Interregio
     Image

In the search, the expanders are not considered as scope. It just links to the previous heading, which is in another expander. Related to #462

MoritzWeber0:
This PR changes the rendering logic of booking pages. Instead of displaying the rendered version of the booking page on the operator page, we do now render it on demand in the operator page context. With some additional changes in this PR, this will fix many issues: 

- Irrelevant sections of booking expanders are no longer rendered to the DOM. This fixes an issue where results where displayed in the search, but were not available on the page itself (see issue description below)
- Images can be rendered out of context now. If you want to render the an image in another page context (for example to render a booking image on an operator page), just write: 
  ```
  ![DB reservation booking](db_reservation.de.webp)
  {page="booking/db-website"}
  ```
- Footnotes of booking expanders are *finally* scoped globally and are no longer rendered locally inside the expander
- The heading map does now properly handle expanders, can be checked via the HeadingsMap extension
- Content of the following expander doesn't count as element from the previous expander in the search (as there was no heading in between, so it counted as element of the last heading of the previous expander)
- Train categories and booking platforms title matches are ranked higher in the search
- If the result targets a booking or train category expander, the page automatically scrolls to it.
- Similar to the already existing behavior with train category expanders, it's now possible to link to booking platforms directly, e. g. https://deploy-preview-856--fipguide.netlify.app/de/operator/oebb/#oebb-website works now.

---

### Issue description

So far, we only hide some booking sections using `display: none` in the CSS of booking expanders. The issue is: Even if hidden, the content is still indexed for the search.

This PR changes the approach to no longer render the irrelevant sections to the DOM by storing the rendered sections in the page store.

Steps to reproduce the search issue: 

1. Search for "Website der Deutschen Bahn" on the German page
2. Select "Reservierungen": 
    grafik
3. The section doesn't exist, no scrolling happens

---

Resolves #887 
Resolves #462


lenderom (dismissed):

assets/js/anchorlinks.js
```diff
@@ -1,5 +1,7 @@
 function initAnchorlinkEventListener() {
-  const anchorLinks = document.querySelectorAll(".a-anchorlink__link");
+  const anchorLinks = document.querySelectorAll(
+    ".a-anchorlink__link, .a-anchorlink__button",
```
therobrob: should we unite the implementations? 
the elements with `.a-anchorlink__link`-class are also `<button>`-Elements..

layouts/_partials/train-category.html
```diff
@@ -122,26 +128,27 @@
...
-          {{- partial "button"
-            (dict
-            "Destination" .additional_information_url
-            "Text" (T "trainCategory.additionalInformation")
-            )
-          -}}
-        {{- end -}}
-      </div>
-    {{- end -}}
+    {{- $content := partial "transformations/increase-headings" (dict "content" .content "offset" 1) -}}
```
therobrob, line 123: do we need this offset? for example here we skip one level:
```
<details>
<h3 style="display:none" id="rj">Railjet (RJ)</h3>
...
<h5 id="rj:reservierungen" data-pagefind-ignore="all">Reservierungen</h5>
</details>
```

layouts/_partials/train-category.html
```diff
@@ -122,26 +128,27 @@
...
+    {{- $content := partial "transformations/transform-expander-headings" (dict "content" $content "id" .id) -}}
+    {{- $content | safeHTML -}}
+    <div data-pagefind-ignore>
```
therobrob: We should add this class to format the buttons well :)

Image

```suggestion
    <div class="o-booking__links" data-pagefind-ignore>
```

therobrob (commented):
Thank you Moritz for your work, these are good improvements! :) i’ve some comments in the code and this further remarks:
* the sources of booking expanders are displayed multiple (for example KW operator page "PKP FIP Information"), is there a way to avoid this?
* i have to click twice on an footnote-backref to scroll to the section
* in @lenderom and my opinion the new anchorlink-button is too big and is shown in the same visual context as the other content-contextual buttons. This doesn’t fit to the mental model of the users. Is there a way to show them like the anchorlink-button next to the headings?
* if there is a heading as the first element in an dialog, we shoud remove the padding-top to format the content well :)

Image

MoritzWeber0:
Thanks for the review, I'll have a look at the suggested changes and will fix them tomorrow.

>     * the sources of booking expanders are displayed multiple (for example KW operator page "PKP FIP Information"), is there a way to avoid this?

This is because of the scoping of the sources in the booking expanders and the separate definition of sources. I have a workaround in mind:
1. We iterate over the sources of the booking expanders
2. If the same entry is used on the main page already, don't use the scoped source and replace it with the main source

I've created an issue for it: #1076

>     * i have to click twice on an footnote-backref to scroll to the section

This is weird, I'll check what is going on there 👀 

>     * in @lenderom and my opinion the new anchorlink-button is too big and is shown in the same visual context as the other content-contextual buttons. This doesn’t fit to the mental model of the users. Is there a way to show them like the anchorlink-button next to the headings?

I do agree and would take out the buttons from this PR and think about it again. I still consider it as useful to have an anchorlink for expanders because I often want to link to a specific train category or booking expander. However, if we place it next to the heading title in the expander summary, we mix it the expansion that would usually happen on click. I can try it out though, maybe it's clear enough from a UI / UX perspective.

>     * if there is a heading as the first element in an dialog, we shoud remove the padding-top to format the content well :)

Will fix it tomorrow as well :)

therobrob:
> I do agree and would take out the buttons from this PR and think about it again. I still consider it as useful to have an anchorlink for expanders because I often want to link to a specific train category or booking expander. However, if we place it next to the heading title in the expander summary, we mix it the expansion that would usually happen on click. I can try it out though, maybe it's clear enough from a UI / UX perspective.

We like the anchor link itself, but we should reconsider its placement and appearance. Perhaps displaying it on a new line as a "ghost button" with a smaller clickable area would be suitable. Displaying it in the expander summary has the disadvantage of overlapping two clickable areas, as you already mentioned.

assets/js/anchorlinks.js
MoritzWeber0: anchorlink implementation is moved to #1077, I added to the command as reminder to the PR description.

layouts/_partials/train-category.html
MoritzWeber0: Good catch, the h3 should have been a h4, I changed it. This is because of the "Fernverkehr" / "Nahverkehr" headings that we have on some pages, for example the ÖBB page: 

image

On the CD page, we have a missing heading level between 2 & 4:

image

This is not optimal from an accessibility perspective (https://www.w3.org/WAI/tutorials/page-structure/headings/). However, this was also the case before of this PR, so I'll create an issue for it: #1078

layouts/_partials/train-category.html
MoritzWeber0: anchorlink implementation is moved to https://github.com/fipguide/fipguide.github.io/pull/1077, I added to the command as reminder to the PR description.

MoritzWeber0:
@therobrob I checked the comments: 

- All anchorlink related changes are now moved to https://github.com/fipguide/fipguide.github.io/pull/1077 for further discussion and alignment
- The footnote-backref double click seems to be bug (browser bug?). I'm also able to reproduce it on the currently deployed fipguide website, so it's not related to this PR. There are a few workarounds available, but they are not beautiful, so I've created an issue for further investigation (#1080)
- I'll fix the skipped heading levels in a separate PR (don't want to blow up this PR), issue is here: #1078
- Same for the merging of the sources, will handle it separately: #1076
- Added consistent handling of expander titles, so all the titles are now underlined on expander hover (this wasn't the case for train category expanders before): https://github.com/fipguide/fipguide.github.io/pull/856/commits/9a572ea482ee2f320aa4495c6a579e548d8d894e
- Fixed the mentioned margin-top for dialogs: https://github.com/fipguide/fipguide.github.io/pull/856/commits/0eac0ad290aa7690a00c2cda97227e54328d7920

MoritzWeber0:
Pushed a few additional fixes I found on the way, for example a few unrendered footnotes, so it's ready again from my side.

therobrob (approved):
LGTM now, thank you! :)

PR #860, opened 2026-04-19
fix: no single booking pages

### LINKED ISSUE #852: Remove single booking pages as direct links

lenderom:
Single booking pages doesn't get rendered as single site anymore.

Fixes: #852


MoritzWeber0:
We should also do a cleanup of the corresponfing layout files and check the CSS rules. I did some of that in https://github.com/fipguide/fipguide.github.io/pull/856 already, but we can also do it separately before.

therobrob (approved):
LGTM, thanks! :)

PR #868, opened 2026-04-21
fix: Handle multi-line HTML in markdown context

MoritzWeber0:
With this change, formatting works in markdown without the need of prettier-ignore.


therobrob (approved):
LGTM, thank you! :)

PR #882, opened 2026-04-23
feat: Change Dartmount Steam Railway to FIP not valid

### LINKED ISSUE #873: Change Dartmount Steam Railway to FIP not valid
FIP is no longer valid.

therobrob:
Resolves #873

### DIFF
--- a/content/operator/gb/index.de.md
+++ b/content/operator/gb/index.de.md
@@ -548,7 +548,7 @@ Reguläre Tickets für Kinder können günstiger sein als Tickets mit FIP-Rabatt
   id="dartmouth-steam-railway"
   title="Dartmouth Steam Railway and River Boat Company"
   type="regional"
-  fip_accepted=partially
+  fip_accepted=false
   reservation_possible=nil
   additional_information_url="https://dartmouthrailriver.co.uk"
 %}}
--- a/content/operator/gb/index.en.md
+++ b/content/operator/gb/index.en.md
@@ -546,7 +546,7 @@ Regular child tickets may be cheaper than tickets with FIP discount.
   id="dartmouth-steam-railway"
   title="Dartmouth Steam Railway and River Boat Company"
   type="regional"
-  fip_accepted=partially
+  fip_accepted=false
   reservation_possible=nil
   additional_information_url="https://dartmouthrailriver.co.uk"
 %}}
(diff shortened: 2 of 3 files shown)


MoritzWeber0:
Is it worth writing a news article about it? It's a change in the FIP acceptance and might be helpful for people who already planned their trips.

therobrob:
> Is it worth writing a news article about it? It's a change in the FIP acceptance and might be helpful for people who already planned their trips.

For other changes we don’t write an news article neither (for example the temporarily required reservations). My opinion: either we write an article for every little change, or we don’t write one at all. :) The user can check changes on the "updated"-mark to be safe.

MoritzWeber0:
> > Is it worth writing a news article about it? It's a change in the FIP acceptance and might be helpful for people who already planned their trips.
> 
> For other changes we don’t write an news article neither (for example the temporarily required reservations). My opinion: either we write an article for every little change, or we don’t write one at all. :) The user can check changes on the "updated"-mark to be safe.

Sounds like we should talk about it in our next alignment: #889

My initial thought is: If there is a change in the FIP rules or regulation - like in this case - we write an news article. 

The addition of the required reservations doesn't seem like a change of rules: The rule that some cross-border trains are reservation-required was already there before, we just added the time periods for convenience. However, it's hard to draw a line.

So I don't want to block this PR - the change makes sense. If we want to add an news article, we can still do it later on.

MoritzWeber0 (approved):

therobrob:
> > > Is it worth writing a news article about it? It's a change in the FIP acceptance and might be helpful for people who already planned their trips.
> > 
> > 
> > For other changes we don’t write an news article neither (for example the temporarily required reservations). My opinion: either we write an article for every little change, or we don’t write one at all. :) The user can check changes on the "updated"-mark to be safe.
> 
> Sounds like we should talk about it in our next alignment: #889
> 
> My initial thought is: If there is a change in the FIP rules or regulation - like in this case - we write an news article.
> 
> The addition of the required reservations doesn't seem like a change of rules: The rule that some cross-border trains are reservation-required was already there before, we just added the time periods for convenience. However, it's hard to draw a line.
> 
> So I don't want to block this PR - the change makes sense. If we want to add an news article, we can still do it later on.

Yeah let’s talk about it in the next meeting! I see there are more perspectives to it and it would be helpful to have some kind of guide ☺️

PR #904, opened 2026-05-01
feat: added travel report from zugpost

therobrob:
Added travel report from Zugpost to german CP page


MoritzWeber0 (dismissed):
Schöner Bericht, danke fürs Ergänzen :)

lenderom:
Adding this great trip report is a really good idea. I appreciate it :D
However what do you think about also adding it also in english and german. I think the selected language should not be relevant for the shown content. However we could add something like (only in german) behind the link. But with modern browser translation this shouldn't be a problem for most users. :)

MoritzWeber0:
> Adding this great trip report is a really good idea. I appreciate it :D
> However what do you think about also adding it also in english and german. I think the selected language should not be relevant for the shown content. However we could add something like (only in german) behind the link. But with modern browser translation this shouldn't be a problem for most users. :)

For Euskotren we didn't add it to the English and French pages, so if we change it, we should also change it there :)

lenderom:
Yeah that's fine for me. I will change it on the Euskotren site and will wait what @therobrob thinks about it :)

therobrob:
> Yeah that's fine for me. I will change it on the Euskotren site and will wait what @therobrob thinks about it :)

That’s okay :) there is an third link on the German renfe page.

lenderom:
I updated this. I also added (nur auf Deutsch) on the german page to keep the translation consistent. Even it's not 100% perfect. But I guess it's okay.

PR #929, opened 2026-05-04
fix: order countries clockwise

### LINKED ISSUE #372: Order of border points / countries
We noticed, that there ic currently no system in which order the countries (from the borders/arrival) on the country page are sorted. For example sorting it alphabetically is not a good idea in my opinion because of the translations in different languages. So I propose to sort it in clockwise starting in the north of the country.

lenderom:
Fixes #372


therobrob (dismissed):
I’ve skimmed through the changes, but I haven’t checked everything in detail. I think it looks fine as is.

therobrob (approved):

PR #935, opened 2026-05-07
fix: add two day rule sheet

### LINKED ISSUE #931: Add 2 day rule translation sheet for print out
We should provide a translation sheet for print out for the two day rule.

https://discord.com/channels/1250522473188032512/1500904171967086673

lenderom:
Fixes #931


MoritzWeber0 (commented):
Should we also make it available as single page in the FIP Guide so that it also scaled on mobile devices? We could still offer the downloadable version.

And we could link it on the FIP validity page too to the operators where we know that the fields are valid for two days.

lenderom:
> Should we also make it available as single page in the FIP Guide so that it also scaled on mobile devices? We could still offer the downloadable version.
> 
> And we could link it on the FIP validity page too to the operators where we know that the fields are valid for two days.

I'm not so sure what the benefit or goal of such an page would be. At the moment I focus on the usecase: "Showing the trains staff a semi-official document to help understanding the rules". I think a scaling responsive site seems less "official" if you know what I mean.

I'm not totallay against such a page, but don't see the benefit at the moment to be honest.

therobrob (approved):
I agree with Lennart that i don't see the benefit for an separate page.

PR #942, opened 2026-05-09
fix: add operators without FIP

### LINKED ISSUE #735: List operators for countries without FIP acceptance
We should list the operators of countries without acceptance somewhere to make them available via the search.

For example, when searching for SJ, the user should be able to see that there is no acceptance, similar to operators without FIP acceptance in countries that have FIP acceptance / are available in the FIP Guide.

lenderom:
Fixes #735

### DIFF
--- a/assets/sass/styles.scss
+++ b/assets/sass/styles.scss
@@ -200,6 +200,10 @@ figure {
   }
 }
 
+.o-list__list--single-column {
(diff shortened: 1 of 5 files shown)


therobrob (approved):
Thank you! 
In #735 Moritz mentioned the index of the page, but i think this is not necessary (as mentioned in your comment).

PR #951, opened 2026-05-12
feat: add IPTC, EXIF and XMP metadata attributes to image elements

### LINKED ISSUE #669: Add IPTC Metadata
We should add the IPTC Metadata information for copyrights of pictures.

This is possible since #665 

> https://github.com/gohugoio/hugo/issues/14549
> https://github.com/gohugoio/hugo/issues/13146

therobrob:
* added EXIF and IPTC metadata support for `artist` and `CopyrightNotice` fields.
* added the information to `figcaption`
* modified intercity direct map – apparently, the metadata doesn't work with WebP, so I uploaded it as a JPG for now.

ToDos: 
* add an note to archetypes.
* check webp support
* our webp images seem to have no copyright information anymore. We have to restore it / add it manually again..


lenderom:
From the review of @therobrob and @lenderom :
It's a nice technical implementation however it doesn't work with .webp. So it would be a step back from our perspective if we change every picture to jpg just to use the metadata. For pictures at the end of the page or news articles isn't a real need for this.

MoritzWeber0:
> From the review of @therobrob and @lenderom :
> It's a nice technical implementation however it doesn't work with .webp. So it would be a step back from our perspective if we change every picture to jpg just to use the metadata. For pictures at the end of the page or news articles isn't a real need for this.

As far as I know, Git also has no proper handling of metadata changes. It's probably treated as a reupload of the file to the repository, which is inefficient.

We thought of sidecar json files before that contain the metadata, maybe that's still an option.

PR #952, opened 2026-05-12
Change in DSB message regarding Regional services.

mikk963s:
Updated the DSB message, GoCo is seizing services in Denmark soon and DSB is looking into overtaking their operations, i will update this message regarding updates to the RE/RA situation.
For now this message is relevant but it might change soon.


mikk963s:
I dont trust autotranslate enough for french and my german is lacking, so if you want to add this info please add the french and german translation 🙏

lenderom:
Thank you for your contribution and insights ♥️

And greetings to Denmark 👋

PR #963, opened 2026-05-25
feat: Add border point map to switzerland

### LINKED ISSUE #930: Visualize Border Points and neighboring countries on map

MoritzWeber0:
Very much work in progress. Only works for Switzerland currently and needs further optimizations.

Resolves #930


MoritzWeber0:
- Make border point markers more visible
- Map decision should be persistent
- Disable map should mention "opt-out"
- Add country overlay

PR #973, opened 2026-05-25
feat: Properly show updates of news articles

MoritzWeber0:
Add a new updated attribute to articles. It should reflect the date of the last major update. This replaces the previous manual handling via updating the original date and the title. The updated date is displayed after the original date as updated tag.

In addition, add an updated partial which can be used to show the updated parts in the news article.


lenderom:
Thanks for the implementation. I think it's a huge improvement for the usability

lenderom (commented):
Thank you for the addition. I think this is a great addition for our users to see which news articles got an update. In the past this wasn't easiely accessible.

I've found some minor improvements:

When having a first glimpse on this PR, I saw badges at the article. However these are not visible anymore:

Image

Also I find the badges a bit distracting / too prominent. For example in the sidebar:

Image

MoritzWeber0:
> When having a first glimpse on this PR, I saw badges at the article. However these are not visible anymore:

Good catch, this was a bug with the updated logic, the page store is only initialized on first render, which didn't happen on the main page. Fixed.

> Also I find the badges a bit distracting / too prominent. For example in the sidebar:

I changed it to normal text in the sidebar, which is less disturbing. For the main page it probably still make sense to generate attention. In the end we want users to recognize that there has been an update.

therobrob:
> I changed it to normal text in the sidebar, which is less disturbing. For the main page it probably still make sense to generate attention. In the end we want users to recognize that there has been an update.

@MoritzWeber0, Lennart and I discussed another design idea for the news-update-tag; i pushed it to the PR (hope that’s okay) and i’m interested in your opinion! :) The tag still generates attention, but integrates better into the page.

[text-box-trim](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-box-trim) is supported by mostly all browsers and will get support by firefox soon.

Bildschirmfoto 2026-07-12 um 21 12 13

lenderom:
- Review von Moritz
- Darkmode Variante
- Sidemenu anpassen (einheitlicher Tag/nicht nur Text)

therobrob (dismissed):
@MoritzWeber0 i’ve added the points discussed. Do you wanna review the adjustments? :)

MoritzWeber0:
> @MoritzWeber0 i’ve added the points discussed. Do you wanna review the adjustments? :)

I noticed that the font-size and paddings for the other tags (train category expanders, booking expanders) were also changed, so I changed them back to the original value and only applied the changes to the news tags. Everything else looks good :)

PR #996, opened 2026-06-13
feat: added information regarding traivelling.com

mfranzke:
I've successfully booked a Eurostar trip through their website, and got the official Eurostar booking codes. Everything went through smoothly, and even the support via Chat was really fast. This way of booking is much easier and instant instead of the other options via email or sending requests that will be replied to on working days, so I would like to promote the offer of this startup.

Update: just recognized the other language versions, I will provide those as well.

and thanks a lot for your great website fipguide.org by the way !


MoritzWeber0:
Hey @mfranzke, thanks for the contribution. I can perhaps explain in more detail why we haven't yet included the traivelling website as a separate booking expander: The other Eurostar booking platforms usually verify the FIP ID card at the time of booking. However, traivelling doesn't ask for this. Nevertheless, we are of course aware that many people are currently using this booking option. We also receive this feedback from the community. Therefore, I would like to discuss this again with the core team to see if we want to include it as a booking expander or how we want to handle it 😊

wouterdedroog:
> The other Eurostar booking platforms usually verify the FIP ID card at the time of booking

For what it's worth: this doesn't apply for NS International either, at least when booking through the phone.

MoritzWeber0:
Hey @mfranzke, some good news from my side. We decided to add the traivelling website as booking expander. I went ahead and implemented + translated it. Feel free to have a look at it.

PS: The anchorlink to the booking expander doesn't work yet. It relies on #1077.

lenderom (approved):

therobrob (approved):

PR #1012, opened 2026-06-20
fix: Close overlay on resize on mobile

### LINKED ISSUE #985: error in the navigation menu (mobile use)
When using the app on a mobile device, an error occurs when opening the navigation menu. 
Switching between portrait and landscape mode causes the menu to become unresponsive, requiring the page to be reloaded. In landscape mode, the desktop navigation is used, which probably causes this error.

Precondition: The smartphone is in portrait mode.
Steps taken:
- open the FIP Guide in any mobile browser (Safari and Chrome tested on iOS)
- open the hamburger menu
- With the menu open, rotate the device to landscape mode

The error then occurs and the page must be reloaded.

MoritzWeber0:
Resolves #985


therobrob (approved):
The error doesn’t occur anymore. There is another unwanted behaviour while using the website in landscape mode. I’ll describe it in a separate issue. :) => #1037

PR #1014, opened 2026-06-20
feat: Add automatic dropdown positioning

### LINKED ISSUE #1013: Navigation dropdown overflows on mobile in landscape mode
Image

It overflows vertically and horizontally.

MoritzWeber0:
Use the new [anchor](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/anchor-name) and [position-try](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@position-try) systems in CSS, available since January in all major browsers to automatically position the dropdown menu.

image

For horizontal positioning, automatically shift the dropdown menu to the left if there is not enough space on the screen. For vertical positioning, use scrolling to handle the overflow.

Resolves #1013


lenderom:
Hi @MoritzWeber0 thank you for your implemenation :)
Is there a way to hide the scrollbar in the menu?
image
(I guess it only appears on smaller screens)

MoritzWeber0:
> Hi @MoritzWeber0 thank you for your implemenation :) Is there a way to hide the scrollbar in the menu? 

It should only appear if the scrollbar is necessary and the content would overflow otherwise, which does happen on browser viewports with < 520px height. There are workarounds to hide scrollbars, but usually they are not really recommended as it also overwrites the system preferences regarding scrollbars. The default on Windows is to show the scrollbar, while on Android or macOS this isn't the case.

assets/sass/dropdown.scss
```diff
@@ -123,23 +128,29 @@
     border: var(--border-visible);
     animation-duration: 0.2s;
     animation-timing-function: ease;
+    max-height: 20rem;
```
therobrob, line 132: Is this an fallback? Do we have an element without modifier? Why `20rem`? It seems somewhat random.

therobrob (commented):
Thank you Moritz – this is a cool new feature; it's just a bit complicated to figure out at first. :) 

**One question:** the language switcher has `o-dropdown__menu--above` AND `o-dropdown__menu--right` at once. Is this correct? The properties are conflicting.

Unfortunately #1037 isn‘t solved by this PR, but we can address it in a separate PR.

layouts/_partials/fip-validity/issuer-dropdown.html
therobrob, line 1: Should we change these to `.id` here as well, just like in nav-dropdown partial?

assets/sass/dropdown.scss
MoritzWeber0: The only place where it might get relevant is the FIP validity dropdown. And the 20rem was just a wild guess, but I wanted to add some fixed height as proposed here: https://github.com/fipguide/fipguide.github.io/pull/827#issuecomment-4231932573

MoritzWeber0:
> **One question:** the language switcher has `o-dropdown__menu--above` AND `o-dropdown__menu--right` at once. Is this correct? The properties are conflicting.

I'll check it later.

> Unfortunately #1037 isn‘t solved by this PR, but we can address it in a separate PR.

👍

layouts/_partials/fip-validity/issuer-dropdown.html
MoritzWeber0: Since the ids are just the identifiers of the operators here, the risk of the same id used somewhere else might be higher. But since the dropdown is the only place where the anchors are used, I'd be fine with it.

Or we use the same as the id of the element, in this case "{{ $id }}-button"? The is should be unique anyway. What do you think?

MoritzWeber0:
> **One question:** the language switcher has `o-dropdown__menu--above` AND `o-dropdown__menu--right` at once. Is this correct? The properties are conflicting.

It's correct that the language switcher has both. One does define the horizontal positioning and one does the vertical alignment (`--right` means it's aligned with the right side of the button).

What wasn't correct is that the `--above` and `--below` descriptors defined rules (esp. `left:`) that shouldn't belong there. I fixed it.

therobrob:
> It's correct that the language switcher has both. One does define the horizontal positioning and one does the vertical alignment (`--right` means it's aligned with the right side of the button).

Okay got it; but why does the navbar-dropdown-menu just have `o-dropdown__menu--below`, but not `o-dropdown__menu--left`? Isn’t it somehow inconstistent?

layouts/_partials/fip-validity/issuer-dropdown.html
therobrob: Maybe we're talking at cross-purposes.
In line 1 we already initialize `$id := .id`, so in L 12 and 30 we can use `.id` instead, couldn’t we?

MoritzWeber0: Ah, thanks for the explanation. I got it and it makes sense. I'll change it to `.id`. There was another `$type` that was unused and could be removed, so I just did it.

MoritzWeber0:
> > It's correct that the language switcher has both. One does define the horizontal positioning and one does the vertical alignment (`--right` means it's aligned with the right side of the button).
> 
> Okay got it; but why does the navbar-dropdown-menu just have `o-dropdown__menu--below`, but not `o-dropdown__menu--left`? Isn’t it somehow inconstistent?

That's fair. `left` was the default and the other options got added later on, but I've added `--left` for consistency now :)

therobrob (approved):

PR #1016, opened 2026-06-20
fix: Update children age limits and rules

### LINKED ISSUE #245: Easier discount overview
In #239 @antvgr changed to discount overview to this formatting:

```
## Discounts

- Children under 4 years: free (no assigned seat)
- Children under 12 years: 40% discount
- From 12 years: full fare
```

I think this gives a better overview then a full text. But I didn't want to change it without voting.

### LINKED ISSUE #685: Add tariff age limits for children
In the FIP contract there is a list of age limits for children. We should add the information to the website.

### LINKED ISSUE #750: Clarify meaning of ages in discount sections
When we talk about ages in the discount section, we should make sure if we include the age or not. Especially in German, that can be unclear quickly. For example GySEV:

> Bei öffentlichen Tarifen fahren Kinder bis 6 Jahren kostenlos. Kinder bis 14 Jahre können 50% Ermäßigung auf den Erwachsenenfahrpreis erhalten. Personen ab 14 Jahre zahlen den Erwachsenenpreis.

Children under 6 travel for free, so when you are aged 6, you can't travel for free anymore. 

Better formulations are "Kinder unter 6", "bis einschließlich 5 Jahren" or "bis zum vollendeten 6. Lebensjahr".

### LINKED ISSUE #920: Compare children discount vs FIP discount
On some pages we have information about the discounts for children. These discounts only apply for the public fare, not the FIP fare.

For example on the DB page: 
> For public fares, [...] children between 6 and 14 years receive a 50% discount on the adult fare.

We should check for each operator: Which discount makes more sense for children? The public fare discount or the 50% from FIP?

For DB, the public discount usually is better because it also applies on savers fares and additional discounts like BahnCard can be used.

MoritzWeber0:
Update all children sections. The section always indicated the discount on the public fare as there is no additional children discount on FIP fares.

Many pages also had outdated age limits, I tried to research the current limits for all operators using their websites. Only for the polish operators I couldn't a public information, so I used the information from RDG.

Resolves #920 
Resolves #685 
Resolves #750
Resolves #239
Resolves #245

- [x] Add Seat61 overview page to FAQ: https://www.seat61.com/european-train-travel-with-children.htm
- [x] Update National Rail children section


MoritzWeber0:
How should we handle these cases? 

- The FIP discount for children is higher than the public fare. Should we still list the public fare? 
- The public fare is cheaper than the FIP fare? Should we still list the FIP fare? 
- Both fares (public and FIP) give 50 % discount. Which one should be mention? Both?

I'm also open to any ideas how to visualize it. I thought about something like this, but the FIP discount is always the same (independent of the age), so the visualization doesn't make much sense:
image

therobrob:
I’ll add my thoughts tomorrow :)
There is another issue regarding the discounts: #245
I think, we can close it, can’t we? :)

content/operator/bls/index.de.md
```diff
@@ -139,7 +139,11 @@ Im Zug ist kein Kauf von FIP-Tickets möglich. Es wird eine Strafgebühr erhoben
 
 ## Ermäßigungen
 
-Für reguläre Fahrten reisen Kinder bis einschließlich 5 Jahre kostenlos. Kinder bis einschließlich 15 Jahre erhalten eine Ermäßigung von 50 % auf den Erwachsenenfahrpreis. Personen ab 16 Jahren zahlen den regulären Erwachsenentarif.[^1]
+Kinder bis einschließlich 5 Jahre reisen kostenlos. Ab 6 Jahren gilt der reguläre FIP-Erwachsenentarif.[^1]
```
lenderom, line 142: I'm a bit unsure about the wording. Do you think we should also mention the 50% discount regardless of FIP. Because als you wrote by yourself in https://github.com/fipguide/fipguide.github.io/issues/920: "the public discount usually is better because it also applies on savers fares and additional discounts like BahnCard can be used."

But maybe that's a point for a little table view you mentioned

MoritzWeber0:
- Only show public fares, but add a note that there is no additional discount for FIP fares.
- Only age categories
- Other discounts like https://www.fipguide.org/en/operator/zssk/#free-travel-for-students-and-seniors-without-fip

therobrob (approved):
I‘ve scrolled through the changes once again, for me this looks fine! Thank you for the good work, Moritz!! :)

PR #1050, opened 2026-07-01
feat: Add filters for booking platforms

MoritzWeber0:


MoritzWeber0:
Update CSS classes and background color

PR #1058, opened 2026-07-02
feat: Add croatia

tintin94230:
Started to work on Croatia as a test ...


MoritzWeber0:
Hey @tintin94230, first of all thanks for the contribution! :)

I saw that the Hugo build was failing, so I applied and pushed two small fixes, hope it helps you to get it running. Since the PR is work-in-progress, I've changed the status to draft. Would you let us know how your plans are with the PR? Do you plan to add additional details or should we take it from here?

PR #1059, opened 2026-07-02
feat: add initial github codespaces config

lenderom:
- add initial github codespaces config

With the free tier we get 120 core-hours per person per month. With the 2-core machine that's 60 hours per month or 30 hours with the 4-core machine. Codespaces can be created here: https://github.com/codespaces

I would recommend stopping the Codespace after a coding session manually.

The first build takes some time. But the Codespace can be kept kind of persistent/reusable when it's not deleted.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
+{
+	"name": "Ubuntu",
+	"image": "mcr.microsoft.com/devcontainers/base:resolute",
+	"features": {
+		"ghcr.io/stu-bell/devcontainer-features/homebrew:1": {
+			"brewPackage": "hugo"
+		},
```
MoritzWeber0, line 7: There seems to be an official devcontainer feature for hugo, but I haven't tested it: https://github.com/devcontainers/features/tree/main/src/hugo

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
...
+		"ghcr.io/devcontainers/features/git-lfs:1": {},
+		"ghcr.io/stu-bell/devcontainer-features/node:0": {
+			"min_node_version": "22.12"
+		},
```
MoritzWeber0, line 11: Same for node, there seems to be an official feature too: https://github.com/devcontainers/features/tree/main/src/node

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,15 @@
...
+		"ghcr.io/devcontainers/features/python:1": {}
+	},
+	"postCreateCommand": "npm install && brew install prek && prek install -f"
```
MoritzWeber0, line 14: Official feature for prek: https://github.com/devcontainers-extra/features/tree/main/src/prek

.devcontainer/devcontainer.json
lenderom: The problem is, it's using an outdated Hugo version. So try for yourself, but I've encountered problems with the language implementation. Brew makes sure we have an up-to-date version of Hugo.

.devcontainer/devcontainer.json
lenderom: Feel free to try it out. Several ways lead to Rome :D

.devcontainer/devcontainer.json
lenderom: The "problem" is. As far as I know this are all standalone dev containers. I'm not sure on how to combine them 😅

lenderom: So feel free to try it out. I'm not motivated to invest more time in this feature 🙈

lenderom:
We should add a entry to our Contribution Guide about the Github Codespaces Usage

lenderom:
@MoritzWeber0 would you like to invest more time in this feature? Otherwise I would like to merge it and test it out in real usecases.

MoritzWeber0 (approved):
Added the prettier plugin for VSCode, automatic port forwarding + switched to the official features (with fixed hugo version). Tested `hugo server` and `prek run -a` and it works, however hugo is really slow (first build took 5 minutes)

MoritzWeber0:
@lenderom Feel free to merge the PR if my changes are fine for you :)

PR #1084, opened 2026-07-13
feat: add prefers-reduced-motion

### LINKED ISSUE #754: Consider `prefers-reduced-motion` CSS media property
For accessibility reasons, we can provide a version with less animations using the `prefers-reduced-motion` CSS property: https://developer.mozilla.org/de/docs/Web/CSS/Reference/At-rules/@media/prefers-reduced-motion

therobrob:
added prefers-reduced-motion to:
- scroll-behaviour
- animations
- transitions

### DIFF
--- a/assets/sass/anchorlink.scss
+++ b/assets/sass/anchorlink.scss
@@ -57,6 +57,10 @@
     background-color 0.3s ease,
     opacity 0.3s ease;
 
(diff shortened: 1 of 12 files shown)


MoritzWeber0 (approved):
Thanks!

PR #1091, opened 2026-07-14
feat: add basic gallery functionality

### LINKED ISSUE #622: Slideshow for decorative content images
For now we can’t display more than one image per page. It would be cool to add a slideshow to display a variety of train-images 🚃❤️
we could eg use [splide](https://splidejs.com/) (my team uses it, because of a11y) or a self coded solution.

therobrob:
Added gallery feature:
- added imageGallery partial and css
- modified helper/contentImage.html
- adjusted other usages of helper/contentImage.html
- modified operator/single.html for gallery feature
- added second image for ŠZ page as an example


therobrob:
hide scrollbar

assets/js/imageGallery.js
```diff
@@ -0,0 +1,42 @@
...
+
+// There is an problem with the scroll-snap on mobile devices, so we need to add the gap between the items when scrolling to make it work properly.
+const galleryGap = 15;
+
+const getItemWidth = () =>
+  galleryPictures.children[0]?.getBoundingClientRect().width ??
+  galleryPictures.clientWidth;
+
+const updateButtonState = () => {
+  console.log("updateButtonState");
```
MoritzWeber0, line 15:
```suggestion
```

There is one left-over `console.log` statement, we should remove it.

assets/sass/imageGallery.scss
```diff
@@ -0,0 +1,47 @@
...
+  margin: 0;
+  padding: 0;
+  gap: 1.5rem;
+  overflow-x: auto;
+  scroll-snap-type: x mandatory;
+  scroll-behavior: smooth;
+  list-style: none;
+  border-radius: var(--border-radius-l);
+
+  -ms-overflow-style: none;
```
MoritzWeber0, line 18: This is just for Internet explorer support, there are a few other places where we don't have internet explorer support, so we could probably just say that we don't support Internet explorer and safely remove it :)

layouts/_partials/helper/contentImage.html
MoritzWeber0, line 1: What do you think about explicitly adding the list of filenames as a front matter attribute?

Something like:

```yaml
contentImages:
  - image.webp
  - image2.webp
```

It would make it a bit more explicit how to add new images without the assumed `image` prefix and would enable CMS support. And we could handle the sorting via the ordered list. I just want to hear your thoughts, we can also happily dismiss the proposal or track it as a separate issue.

MoritzWeber0 (commented):
LGTM, just three small comments.

layouts/_partials/helper/contentImage.html
therobrob: Thanks! I like the idea, but I would track it in a separate issue. :)
https://github.com/fipguide/fipguide.github.io/issues/1194

PR #1115, opened 2026-08-01
feat: update Regionalverkehre Start GmbH

### LINKED ISSUE #635: Start GmbH will be integrated into DB Regio AG
https://www.ndr.de/nachrichten/niedersachsen/lueneburg_heide_unterelbe/db-regio-uebernimmt-start-gmbh-zuege-sollen-weiter-fahren,bahn-1054.html

DB Regio AG takes over Start GmbH. The existing routes will be operated by DB Regio from 1st of August 2026 on. It's unclear if FIP is valid on these DB Regio routes, but probably it won't be accepted.

therobrob:


MoritzWeber0 (approved):
Thanks for the update, looks good to me! We can already merge it, but I wonder if we should add an news article since these routes have a touristic relevance due to the connection with the Harzer Schmalspurbahn :)

PR #1192, opened 2026-09-16
Update to the use of FIP facilities with Scandlines/DB Regio Stuttgart

calvin-zeu:
* Add information on the resp. country pages that the ferries to / from `germany` and `denmark` can't be used with FIP facilities;
* Remove the unofficial information from `dsb` operator page that Scandlines may be used with FIP facilities. An email from Scandlines is [available](https://github.com/user-attachments/assets/d62bbcc4-72c5-49c0-9db9-77c5a6298754);
* Remove information about DB Regio Stuttgart from `db` operator page as DB Regio Stuttgart GmbH is fully integrated into DB Regio AG, Region Baden-Württemberg;
* Add more information to the Bordrestaurant and ICE Portal of operator `db`.


lenderom (approved):
Thanks @calvin-zeu for this great edit. This is really helpful <3

PR #1203, opened 2026-09-20
fix: add connections section to en and fr pages

### LINKED ISSUE #1175: English & French NS page is missing connection information
The english and french versions of the NS page are missing the connection information that is available in the german version below the Anschlüsse section.

therobrob:
- added the `Anschlüsse`-section to english and french pages
- corrected en-dashes
- corrected ns-borderpoints according to the `germany`-page


lenderom (approved):
❤️🎉

